PR #40, opened 2026-05-04
split resume media into dedicated transcript and recording endpoints

sawera-fia:
This PR refactors the recruitment module's media retrieval logic. By splitting the combined media endpoint into two dedicated routes.
Key Changes:
New API Endpoints:
GET /by-resume/{resume_id}/transcript: Retrieves the full conversation transcript and session status.
GET /by-resume/{resume_id}/recording: Retrieves the audio recording URL and session identifiers.
Centralized Resolution Logic:
Introduced a private helper function _get_latest_call_session_for_resume to handle the multi-step lookup (Resume -> Interview -> Call Session).
Robust Filtering for Multiple Attempts:
Modified the interview retrieval logic to automatically skip empty "In Progress" records. The system now searches for the most recent interview record that contains a valid call session, ensuring users always see the latest successful interaction even if multiple scheduling attempts exist.
Data Models:
Implemented ResumeInterviewTranscriptResponse and ResumeInterviewRecordingResponse to ensure strict typing and optimized payload delivery.

### DIFF
--- a/app/routers/resume_interviews.py
+++ b/app/routers/resume_interviews.py
@@ -8,6 +8,7 @@ from uuid import UUID
 from fastapi import APIRouter, Depends, HTTPException, Query, status
 from sqlalchemy import or_
 from sqlalchemy.orm import Session
+from app.core.logger import logger
 
 from app.api.deps import get_db, require_admin_or_owner, require_member_or_admin
 from app.models.job_description import JobDescription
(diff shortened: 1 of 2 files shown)


mubeenhussain8:
1. Do not remove the old endpoint when you add a new one instead deprecate it first or align with frontend.
2. Check the calendar endpoint first and follow the same style/pattern.
3. Instead of lots of “latest” filtering logic, use a clear interview_id or session_id to fetch transcript and recording.

PR #71, opened 2026-05-19
Feature/pii redaction logging

Aliict:
app/core/pii_redactor.py — new redact_pii(value) utility with compiled regex patterns covering emails, phone numbers (E.164 + US formats), credit/debit card numbers (13–19 digits), SSNs, bank/account numbers, and honorific-prefixed names. Handles nested dict, list, tuple, and bytes recursively with a depth-20 guard.
app/core/logger.py — added _PiiRedactionFilter(logging.Filter) attached to the root logger and all its handlers inside setup_logging(). Every log record's msg and args are passed through redact_pii before any handler emits the record. No changes needed in service or router code — existing logger.* calls are automatically scrubbed.
app/main.py — added three global @app.exception_handler hooks:
HTTPException → passes exc.detail through redact_pii before responding
RequestValidationError → sanitises the Pydantic error list
Exception (catch-all) → logs with exc_info (redacted by the filter), returns a generic 500 message so raw exception text never reaches the client


app/core/pii_redactor.py
```diff
@@ -0,0 +1,254 @@
...
+    }
+)
+
+# Header values passed through unchanged (timestamps / correlation ids — not PII).
+_PASS_THROUGH_HEADER_KEYS: frozenset[str] = frozenset(
+    {"x-request-start", "x-request-id"},
+)
+
+# Query-string secrets in URLs logged by httpx/urllib3 (Trello, OAuth, etc.).
+_URL_SECRET_PARAM_RE = re.compile(
```
codewithwaleed, line 43: **[REQUIRED] SSN pattern `\b\d{9}\b` is extremely broad — will false-positive on UUIDs, Twilio SIDs, and innocent 9-digit numbers**

```python
re.compile(r"\b\d{3}[- ]\d{2}[- ]\d{4}\b|\b\d{9}\b"),
```

The `\b\d{9}\b` branch matches any 9-digit sequence. This will redact:
- Twilio call SIDs (contain 9-digit substrings)
- ZIP+4 codes, timestamps, or numeric IDs that happen to be 9 digits
- Log lines like `call_duration=123456789`

This breaks observability — redacted Twilio SIDs make call logs undebuggable.

**Fix:** Keep only the hyphenated SSN format which has much lower false-positive rate:
```python
re.compile(r"\b\d{3}[- ]\d{2}[- ]\d{4}\b"),  # SSN: ###-##-#### only
```

The bare 9-digit form is rarely useful and the false-positive cost is too high.

app/core/logger.py
```diff
@@ -2,40 +2,94 @@
...
+
+def _redact_log_value(value: object) -> object:
+    """Redact a single log-format arg (exceptions must become redacted strings)."""
+    from app.core.pii_redactor import redact_pii
+
+    if isinstance(value, BaseException):
+        return redact_pii(str(value))
+    return redact_pii(value)
+
+
```
codewithwaleed, line 23: **[WARNING] `_PiiRedactionFilter.filter` silently swallows exceptions from `redact_pii`**

```python
try:
    record.args = _redact_log_args(record.args)
except Exception:
    pass
```

If `redact_pii` raises (e.g. regex engine error on a pathological string), the original unredacted args are left in `record.args` and emitted to the log unchanged. This is a silent PII leak.

**Fix:** On exception, replace args with a safe placeholder rather than leaving them untouched:
```python
except Exception as exc:
    record.args = ("[PII redaction error — args suppressed]",)
    record.msg = record.msg  # keep message structure but args are gone
```

app/core/pii_redactor.py
```diff
@@ -0,0 +1,254 @@
...
+    (
+        REDACTED,
+        re.compile(r"[a-zA-Z0-9._%+\-]+@[a-zA-Z0-9.\-]+\.[a-zA-Z]{2,}", re.IGNORECASE),
+    ),
+    # Stripe Checkout hosted URLs (contain session secrets in path/query)
+    (
+        REDACTED,
+        re.compile(r"https://checkout\.stripe\.com/[^\s\"'\\]+", re.IGNORECASE),
+    ),
+    # Stripe Checkout session ids (cs_test_… / cs_live_…)
```
codewithwaleed, line 75: **[WARNING] Phone number pattern `\+?[0-9]{10,15}\b` will match Stripe session IDs and API keys**

The E.164 pattern:
```python
re.compile(r"\+?[0-9]{10,15}\b"),
```

Will match any 10–15 digit numeric sequence — including Stripe payment intent IDs (`pi_3...`), timestamp microseconds in structured logs, and sequential IDs. This adds to the false-positive risk alongside the SSN issue.

**Fix:** Anchor the pattern more tightly by requiring it to appear standalone or after a non-digit boundary, and exclude patterns that follow `pi_`, `ch_`, `cs_` etc. At minimum, test the pattern against a sample of your actual log output before enabling in production.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #71: PII Redaction + Centralised Logging

**Overall: Strong security foundation.** The compiled regex patterns, depth-guarded recursion, and filter-based approach (no call-site changes needed) are all well-designed. One required fix and two warnings.

app/core/pii_redactor.py
Aliict: Removed overly broad 9-digit SSN matching pattern from regex to reduce false positives and preserve observability in logs, keeping only the strict hyphenated SSN format.

app/core/logger.py
Aliict: Improved `_PiiRedactionFilter` error handling to prevent silent PII leakage by suppressing log arguments with a safe placeholder when redaction fails instead of emitting unredacted data.

app/core/pii_redactor.py
Aliict: Refined phone number redaction pattern review by addressing overly broad E.164 matching and reducing false positives from structured IDs, with recommendation to tighten validation and test against real logs before production use.

PR #72, opened 2026-05-20
Feature/middleware error handling

Aliict:
Standardised error envelope — every error response across the entire app (HTTP exceptions, validation errors, 404s, unhandled 500s, 401s, 413s) now returns { "error": { "code": "<snake_case>", "message": "…", "requestId": "<nanoid>" } }. The old flat { error_code, message, status_code } shape is gone. A new RequestIdMiddleware generates a nanoid per request and threads it through request.state, response headers (X-Request-ID), and every error body.

Hardened middleware stack — five changes made the stack production-safe: CORS origins now come from the ALLOWED_ORIGINS env var (no more ["*"] hardcoded); a new BodyLimitMiddleware rejects requests over 10 MB with HTTP 413; PiiLoggingMiddleware was extended to log METHOD PATH STATUS DURATIONms requestId=… at INFO on every request; the @on_event startup/shutdown hooks were replaced with the modern lifespan context manager that also handles SIGTERM; and the Dockerfile CMD now honours $PORT at container start time.


app/core/exception_handlers.py
```diff
@@ -33,29 +57,41 @@ async def validation_exception_handler(
         content=build_api_error_payload(
             422,
```
codewithwaleed, line 59: **[REQUIRED] `not_found_handler` registered for status 404 AND `http_exception_handler` handles `HTTPException` — they will conflict for 404s**

From the comment:
> `http_exception_handler` already covers it; `not_found_handler` is the explicit hook for the 404 status specifically.

In Starlette, when both `app.add_exception_handler(HTTPException, ...)` and `app.add_exception_handler(404, ...)` are registered, **the integer-keyed handler (404) takes priority over the exception-class handler for that status code**. This means 404s raised by route handlers via `raise HTTPException(status_code=404)` will hit `not_found_handler` instead of `http_exception_handler` — losing the `exc.detail` message and always returning `"Route not found"`.

**Fix:** Remove `not_found_handler` and the `app.add_exception_handler(404, not_found_handler)` line. The `http_exception_handler` already handles all `HTTPException` instances including 404 with correct detail message and `X-Request-ID` header.

app/core/config.py
```diff
@@ -21,6 +21,11 @@ class Settings(BaseSettings):
     HOST: str = "0.0.0.0"
     PORT: int = 8000
     DEBUG: bool = True
+    APP_VERSION: str = "1.0.0"
+
+    # CORS — comma-separated list of allowed origins.
```
codewithwaleed, line 26: **[WARNING] `DEBUG: bool = True` default — production deployments must explicitly set `DEBUG=False`**

`DEBUG=True` enables verbose logging that may expose stack traces and internal paths. Since `pydantic-settings` reads from env vars, this is safe as long as `DEBUG=False` is set in staging/production environments. However, if that env var is ever missing (misconfigured deploy), the app silently runs in debug mode in production.

**Fix:** Change the default to `False` and require it to be explicitly opted into for development:
```python
DEBUG: bool = False
```

Update `.env.example` with `DEBUG=True` for developers.

Dockerfile
```diff
@@ -16,8 +16,9 @@ RUN pip install --no-cache-dir -r requirements.txt
 # Copy the rest of the application's code into the container
 COPY . /app
 
-# Expose port 8000 to the outside world
-EXPOSE 8000
+# PORT defaults to 8000; override via environment variable at runtime.
+ENV PORT=8000
```
codewithwaleed, line 20: **[WARNING] Dockerfile missing final newline — minor but causes `git diff` noise**

The file ends without a trailing newline (`\ No newline at end of file`). While not a functional issue, POSIX requires text files to end with a newline and many tools (git, linters) warn about this. Add a newline at the end of the `CMD` line.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #72: Standardised Error Envelope + Hardened Middleware Stack

**Overall: Excellent infrastructure PR.** The unified error envelope, `X-Request-ID` threading, lifespan migration, CORS from env var, and body size limit are all production-grade improvements. One required fix and two warnings.

app/core/exception_handlers.py
Aliict: Removed conflicting 404-specific exception handler to ensure `http_exception_handler` consistently handles all `HTTPException` cases, preserving error details and avoiding response override issues.

app/core/config.py
Aliict: Changed `DEBUG` default to `False` to ensure production-safe behavior and require explicit opt-in for debug mode, preventing accidental exposure of sensitive debug information in deployments.

Dockerfile
Aliict: Added missing trailing newline at end of Dockerfile to comply with POSIX standards and eliminate unnecessary git diff warnings.

PR #73, opened 2026-05-21
Feature/invite rate limiter

Aliict:
Removed the old invite endpoint and per-routelogin_rate_limit; added tests for the new workspace invite endpoints.
Refactored invite flow: updated invite model/status handling, wired SendGrid viaemail_service, and rolled back invite on email failure.


app/api/api_v1/endpoints/user.py
```diff
@@ -17,7 +17,6 @@
 from app.core.security import is_token_expired, verify_token
 from app.services.email_service import email_service
 from app.utils.response import create_success_response
-from app.utils.rate_limiter import login_rate_limit
 from datetime import datetime, timezone
 from app.services.role_service import get_user_role_in_tenant
```
codewithwaleed, line 21: **[REQUIRED] `@login_rate_limit()` decorator removed with no replacement at the route level**

The per-route `@login_rate_limit()` on `POST /login` was removed, and `API_RATE_LIMIT` was reduced from 1000 → 60 req/min as the global rate limit. However, the global `RateLimitMiddleware` (added in PR #72) applies the same 60/min budget to **all** authenticated endpoints equally — login gets the same allowance as fetching a call log.

Login endpoints are high-value brute-force targets and should have a **stricter, dedicated limit** (e.g. 10 req/min per IP before auth) regardless of the global limit. Removing the per-route decorator without a compensating tighter rule is a regression in brute-force protection.

**Fix:** Re-add a dedicated login rate limit, either via a tighter Redis key specifically for `/login`, or by restoring the decorator with a lower cap.

app/api/api_v1/endpoints/workspace_invites.py
```diff
@@ -0,0 +1,102 @@
...
+    body: InviteCreate,
+    admin: User = Depends(require_admin),
+    db: Session = Depends(get_db),
+) -> SuccessResponse[InviteOut]:
+    tenant_id: uuid.UUID = admin.current_tenant_id
+
+    tenant = db.query(Tenant).filter(Tenant.id == tenant_id).first()
+    if not tenant:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Tenant not found")
+
```
codewithwaleed, line 39: **[REQUIRED] Invite token uses `uuid.uuid4()` — predictable format, weaker than `secrets.token_urlsafe`**

```python
token = str(uuid.uuid4())
```

UUID v4 is random but its format (`xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx`) is well-known and can be pattern-matched. The old implementation used `secrets.token_urlsafe(32)` which provides 256 bits of cryptographic randomness in a URL-safe encoding.

**Fix:** Restore the cryptographically stronger token:
```python
import secrets
token = secrets.token_urlsafe(32)
```

app/middleware/rate_limit_middleware.py
```diff
@@ -0,0 +1,209 @@
+"""
+Global sliding-window rate limiter middleware.
+
+Enforces API_RATE_LIMIT requests per API_RATE_WINDOW seconds per authenticated
+identity using Redis sorted sets (true sliding window, multi-pod safe).
+
+Identity resolution (first match wins):
+  api_key auth  → SHA-256 of raw x-api-key header
+  jwt auth      → "jwt:{user_id}"
+  fallback      → "ip:{client_host}"
```
codewithwaleed, line 10: **[WARNING] `_SKIP_PREFIXES` whitelist skips rate limiting on `/api/v1/users/` entirely**

```python
_SKIP_PREFIXES = (
    "/api/v1/users/",
    "/api/v1/api-keys/",
    "/api/v1/accept-invite",
)
```

Skipping `/api/v1/users/` means registration (`POST /users`), password reset (`POST /users/forgot-password`), and any other user endpoints are **completely unrate-limited**. These are exactly the endpoints most targeted by bots for account enumeration and credential stuffing.

**Fix:** Be more specific — skip only the paths that genuinely need no limit (e.g. health checks), and apply stricter limits to auth-adjacent endpoints:
```python
_SKIP_EXACT = {"/", "/docs", "/redoc", "/openapi.json", "/api/v1/health"}
# Do NOT skip /api/v1/users/ wholesale
```

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #73: Invite Rate Limiter + Workspace Invite Refactor

**Overall: Good cleanup.** Consolidating invite logic under `/workspace` prefix, normalising status strings to lowercase, and fixing the timezone-aware comparison are all correct improvements. Two required fixes and one warning.

app/api/api_v1/endpoints/user.py
Aliict: Restored dedicated rate limiting for the /login endpoint to ensure stricter per-IP protection against brute-force attempts, independent of the global API rate limit.

app/api/api_v1/endpoints/workspace_invites.py
Aliict: Replaced UUID-based invite token generation with secrets.token_urlsafe(32) to ensure stronger cryptographic randomness and improve security of invite links.

app/middleware/rate_limit_middleware.py
Aliict: Refined rate-limit skip rules by removing broad `/api/v1/users/` prefix exemption and restricting bypass only to safe system endpoints like health and docs to prevent unprotected access to sensitive user-related routes.

PR #74, opened 2026-05-21
feat(database): implement schema v1 updates with soft delete and role…

Aliict:
… enhancements

- Introduce soft delete functionality for users and tenants by adding `deleted_at` columns.
- Add `updated_at` columns to key models (User, Tenant, ApiKey, Invite, RefreshToken) for tracking modifications.
- Implement new roles: `config` for workspace configuration and `readonly` for limited access.
- Update role management to enforce access restrictions based on user roles.
- Enhance database schema with foreign key constraints and composite indexes for improved data integrity.
- Add migration scripts to apply these changes and ensure backward compatibility.

This update strengthens data management and access control within the application.


alembic/versions/20260521_schema_v1_gaps.py
```diff
@@ -0,0 +1,155 @@
...
+down_revision: Union[str, Sequence[str], None] = "20260518_tenant_name_uq"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ------------------------------------------------------------------ tenant
+    op.add_column(
+        "tenant",
+        sa.Column("updated_at", sa.DateTime(timezone=True), nullable=True),
```
codewithwaleed, line 63: **[REQUIRED] `downgrade()` drops the new FK and recreates the old one without `ON DELETE` — but the old constraint name may already be gone**

```python
op.drop_constraint("user_current_tenant_id_fkey", "user", type_="foreignkey")
op.create_foreign_key("user_current_tenant_id_fkey", ...)  # no ondelete
```

PostgreSQL auto-generates FK constraint names based on table + column. If the original name was anything other than `user_current_tenant_id_fkey`, the `drop_constraint` will raise `ProgrammingError` and block the rollback.

**Fix:** Use `_has_constraint` (same pattern as PR #76/#78) before dropping, or use `op.execute(sa.text(...))` with `information_schema` lookup to find the real constraint name dynamically.

alembic/versions/20260521_role_config_readonly.py
```diff
@@ -0,0 +1,65 @@
...
+
+Run:
+    alembic upgrade head
+Revert:
+    alembic downgrade -1
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
```
codewithwaleed, line 22: **[REQUIRED] `downgrade()` deletes roles without checking for existing assignments — cascades silently**

```python
op.execute(sa.text("DELETE FROM role WHERE name IN (\'config\', \'readonly\')"))
```

The comment says "only if no users are assigned to them" but the SQL has no such guard — it deletes unconditionally. If any `user_tenant_association` row references one of these roles, the FK `ON DELETE` behaviour determines what happens (if the FK has no `ondelete`, this will raise a foreign key violation and block the rollback).

**Fix:**
```python
op.execute(sa.text(
    """
    DELETE FROM role
    WHERE name IN (\'config\', \'readonly\')
      AND id NOT IN (SELECT DISTINCT role_id FROM user_tenant_association WHERE role_id IS NOT NULL)
    """
))
```

app/api/deps.py
```diff
@@ -17,13 +17,24 @@
...
+from app.services.role_service import is_admin_in_tenant, get_user_role_in_tenant
 from app.services.role_service import get_user_product_in_tenant
 import uuid
 
 security = HTTPBearer()
 security_optional = HTTPBearer(auto_error=False)
 
+_WRITE_METHODS = frozenset({"POST", "PUT", "PATCH", "DELETE"})
+
+
```
codewithwaleed, line 40: **[WARNING] `_reject_readonly_on_write` is a standalone function — not enforced automatically**

The helper is defined but must be called explicitly by each dependency that needs it. If a developer adds a new `require_*` dependency and forgets to call `_reject_readonly_on_write`, readonly users can mutate data through that endpoint.

**Recommended pattern:** Apply it inside `require_user_tenant` or a new `require_write_access` dependency that wraps `require_user_tenant`, so enforcement is automatic for any route that depends on it:
```python
def require_write_access(request: Request, user: User = Depends(require_user_tenant), db: Session = Depends(get_db)) -> User:
    role = get_user_role_in_tenant(db, user.id, user.current_tenant_id)
    if role:
        _reject_readonly_on_write(request, role.name)
    return user
```

app/api/api_v1/endpoints/user.py
```diff
@@ -115,8 +115,11 @@ def login(login_data: LoginRequest, db: Session = Depends(get_db)):
     """
     # Convert email to lowercase
     login_data.email = login_data.email.lower()
-    # Find user by email
-    user = db.query(User).filter(User.email == login_data.email).first()
+    # Find active (non-deleted) user by email
+    user = db.query(User).filter(
+        User.email == login_data.email,
+        User.deleted_at.is_(None),
+    ).first()
```
codewithwaleed, line 129: **[SUGGESTION] Soft-deleted user query not yet applied to `get_current_user_jwt` JWT lookup**

Multiple auth paths now filter `User.deleted_at.is_(None)` — login, Google login, token refresh, forgot_password — but verify that `get_current_user_jwt` (in `app/api/deps.py`) also excludes soft-deleted users. A deleted user whose JWT has not yet expired could otherwise continue making authenticated requests until token expiry.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #74: Schema v1 Updates (Soft Delete, Roles, FK Constraints)

**Overall: Well-structured schema hardening.** Soft-delete is consistently applied across all auth paths, the role CHECK constraint is a good guardrail, and the FK `ON DELETE SET NULL` is the correct semantics for `user.current_tenant_id`. Two required fixes and one warning.

alembic/versions/20260521_schema_v1_gaps.py
Aliict: Improved Alembic downgrade safety by ensuring the foreign key is dropped only if it exists (or resolved dynamically) to prevent rollback failures due to mismatched constraint names.

alembic/versions/20260521_role_config_readonly.py
Aliict: Fixed Alembic downgrade to safely delete roles only when they are not referenced by any user assignments, preventing accidental constraint violations during rollback.

app/api/api_v1/endpoints/user.py
Aliict: Ensured `get_current_user_jwt` also filters out soft-deleted users (`deleted_at IS NULL`) to prevent access using valid JWTs after account deletion until token expiry.

app/api/deps.py
Aliict: Integrated `_reject_readonly_on_write` into a dedicated `require_write_access` dependency to ensure write restrictions are enforced automatically across all protected routes, preventing accidental omission in future endpoints.

PR #75, opened 2026-05-21
feat(agent-management): add agent ticket fields and CRUD operations

Aliict:
- Introduce new fields for agent management: `status`, `llm_model`, `tts_provider_slug`, `tts_voice_external_id`, `tts_language`, and `encrypted_elevenlabs_api_key` to support the ticket-shaped contract for the `/api/v1/agent` endpoints.
- Implement migration to add these fields to the database schema and relax `created_by` and `updated_by` fields to be nullable for machine-to-machine API key requests.
- Create `AgentRepository` for SQL access and implement CRUD operations for agents, including validation for allowed LLM models and encryption for BYO ElevenLabs keys.
- Enhance agent service logic to handle ticket payloads and updates, ensuring compliance with new field requirements.
- Update agent-related endpoints to utilize the new fields and improve error handling for invalid LLM models.

This update enhances the agent management capabilities and aligns with the new API contract.


app/core/agent_runtime.py
```diff
@@ -0,0 +1,212 @@
...
+        provider_slug="gemini",
+        api_key=None,
+        temperature=0.15,
+        max_tokens=100,
+        used_ticket_llm=False,
+    )
+    if not agent:
+        return default
+
+    temperature = 0.15
```
codewithwaleed, line 91: **[REQUIRED] `_TICKET_TTS_TO_ADAPTER` maps `"rime"" → "google"" with no warning at call time**

```python
"rime": "google",  # no Rime adapter yet — safe telephony fallback
```

When an agent with `tts_provider_slug="rime"` makes a call, it silently uses Google TTS. The caller sees wrong voice and hears nothing that matches what they configured. The debug log fires only at `DEBUG` level — invisible in production where `INFO` is the floor.

**Fix:** Promote to `WARNING` level and include the agent ID so it\'s visible in production logs:
```python
if slug == "rime":
    logger.warning(
        "Agent %s: Rime TTS not yet available — falling back to Google TTS",
        agent.id,
    )
```

This is a tracking issue for the Rime adapter (see PR #77) but the silent fallback is dangerous in the meantime.

alembic/versions/20260522_agent_ticket_fields.py
```diff
@@ -0,0 +1,149 @@
...
+def _has_index(conn, index_name: str) -> bool:
+    return conn.execute(
+        sa.text(
+            "SELECT EXISTS (SELECT 1 FROM pg_indexes WHERE indexname = :ix)"
+        ),
+        {"ix": index_name},
+    ).scalar()
+
+
+def upgrade() -> None:
```
codewithwaleed, line 141: **[REQUIRED] `downgrade()` does not guard `alter_column` with idempotency checks**

The upgrade adds columns with `_has_column` guards (safe to re-run), but downgrade blindly calls:
```python
op.alter_column("agent", "updated_by", nullable=False)
op.alter_column("agent", "created_by", nullable=False)
```

If these columns already have `nullable=False` (e.g. partial downgrade was already run), PostgreSQL will raise an error. Also, if existing rows have `NULL` in `created_by`/`updated_by` (written by M2M requests after the upgrade), restoring `NOT NULL` will fail with a constraint violation.

**Fix:** Add a pre-check in downgrade — either skip the `alter_column` if already `NOT NULL`, or `UPDATE agent SET created_by = \'...\'  WHERE created_by IS NULL` with a placeholder before tightening the constraint.

app/core/agent_runtime.py
```diff
@@ -0,0 +1,212 @@
...
+            pname = (agent.provider.name or "").lower()
+            if "openai" in pname:
+                provider_slug = "openai"
+            elif "groq" in pname:
+                provider_slug = "groq"
+            elif "gemini" in pname or "google" in pname:
+                provider_slug = "gemini"
+        api_key = None
+        if agent.model.api_key:
+            try:
```
codewithwaleed: **[WARNING] `decrypt_api_key` exception swallowed silently — agent runs without API key**

```python
except Exception as exc:
    logger.error("Failed to decrypt legacy model API key: %s", exc)
```

After logging, execution continues and `api_key=None` is returned. The LLM service will attempt to call the provider without an API key, producing a confusing 401 from the provider rather than a clear "BYO key is corrupted" error message to the operator.

**Fix:** Either raise a `RuntimeError` so the call fails fast with a clear message, or surface this as a structured 500 in the call-initiation path, not a silent `None`.

app/core/agent_runtime.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+_TICKET_TTS_TO_ADAPTER: dict[str, str] = {
+    "11labs": "elevenlabs",
+    "11labs_byo": "elevenlabs",
+    "rime": "google",  # no Rime adapter yet — safe telephony fallback
+}
+
+
+@dataclass(frozen=True)
```
codewithwaleed, line 28: **[WARNING] Default LLM hardcoded as `gemini-1.5-flash` in `resolve_llm_runtime`**

```python
model_name="gemini-1.5-flash",
provider_slug="gemini",
```

If Gemini is deprecated or the default model changes, this requires a code change + deploy. Move to `settings.py`:
```python
DEFAULT_LLM_MODEL: str = "gemini-1.5-flash"
DEFAULT_LLM_PROVIDER: str = "gemini"
```

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #75: Agent Ticket Fields & CRUD

**Overall: Solid foundation for the ticket-shaped agent contract.** Migration is idempotent, the repository layer is clean, and the fallback chain in `agent_runtime.py` is well thought-out. Two required fixes and two warnings below.

app/core/agent_runtime.py
Aliict: Upgraded the Rime → Google TTS fallback logging to WARNING level with agent context to make silent provider fallback visible in production and improve traceability until full Rime adapter support is implemented.

app/core/agent_runtime.py
Aliict: Improved `decrypt_api_key` error handling by preventing silent fallback to `None` and ensuring failures surface explicitly so missing or corrupted API keys fail fast with a clear error instead of causing downstream provider 401s.

app/core/agent_runtime.py
Aliict: Moved default LLM configuration from hardcoded values to settings to allow runtime configurability and avoid code changes when switching models or providers.

alembic/versions/20260522_agent_ticket_fields.py
adeelasgher847: @Aliict Is it done?

Aliict: yes

PR #76, opened 2026-05-25
feat(telephony): Twilio provisioning, BYO/SIP external numbers, agent binding

Aliict:
## Summary

Implements Sprint 2 phone number provisioning ticket end-to-end:

- **Twilio number search & atomic purchase** — `GET /api/v1/phone-numbers/search`, `POST /api/v1/phone-numbers/purchase` (persists to DB in one transaction, configures inbound webhook immediately after purchase)
- **BYO/SIP external number registration** — `POST /api/v1/telephony/external` (`provider=external`, SIP credentials encrypted at rest)
- **Agent binding lifecycle** — `POST /api/v1/telephony/bind` (sets `agent.status = ready`), `POST /api/v1/telephony/unbind` (sets `agent.status = pending`)
- **Binding status listing** — `GET /api/v1/phone-numbers` upgraded to return `agent_id`, `agent_name`, `agent_status`, `binding_status` per row; `GET /api/v1/telephony/bindings` for bound-only view
- **Secret Manager** — `app/core/secret_manager.py` routes staging → Twilio test creds (no real purchases), production → GCP Secret Manager, dev → env vars
- **`numberconfiguration` table** — per-number `recording_enabled`, `max_duration_seconds`, `business_hours` JSONB; default row created on every purchase/external registration
- **13 integration tests** — all passing (Twilio HTTP boundary mocked via `unittest.mock.patch`)

## API contract delivered

| Method | Path | Status |
|--------|------|--------|
| GET | `/api/v1/phone-numbers/search?country=AU&type=local&areaCode=02` | ✅ |
| POST | `/api/v1/phone-numbers/purchase` | ✅ |
| POST | `/api/v1/telephony/external` | ✅ |
| POST | `/api/v1/telephony/bind` | ✅ |
| POST | `/api/v1/telephony/unbind` | ✅ |
| GET | `/api/v1/phone-numbers` | ✅ upgraded with binding info |
| GET | `/api/v1/telephony/bindings` | ✅ bound-only view |

## Schema changes

Run `alembic upgrade head` — migration is idempotent (`_has_column`/`_has_table` guards).

| Table | Change |
|-------|--------|
| `phonenumber` | Add `provider` (default `twilio`), `sip_username`, `sip_password` (encrypted) |
| `numberconfiguration` | New table — FK to `phonenumber.id`, `recording_enabled`, `max_duration_seconds`, `business_hours` JSONB |

## Design decisions

**`assistant_id` → `agent_id`**: DB column stays `assistant_id` for backward compatibility with `voice_call_service`, `scheduled_call_service`, and `appointment_follow_up_service`. All new endpoints use `agent_id` in request/response bodies via Pydantic `AliasChoices`. Service layer maps between them — no data migration needed.

**Agent status `pending`/`ready`**: Added alongside existing `active`/`inactive`/`draft`. Column is `String(20)` — no DDL enum change. Bind → `ready`; unbind → `pending`.

**Secret Manager**: `get_twilio_credentials()` is called inside `TwilioService.get_client()` so all Twilio operations automatically use correct credentials. Staging is enforced at the service level — test creds are returned regardless of what's in env.

## Not in scope

`POST /api/v1/calls/inbound` — Sprint 4. Existing `/api/v1/voice/incoming` webhook is unchanged.

## Test plan

- [ ] `python3 -m pytest tests/api/test_phone_provisioning.py tests/api/test_agents.py -v` → 40 tests pass
- [ ] `alembic upgrade head` on staging DB
- [ ] Set `ENVIRONMENT=staging` + `TWILIO_TEST_ACCOUNT_SID`/`TWILIO_TEST_AUTH_TOKEN` → purchase uses test creds, no real number charged
- [ ] Set `GCP_PROJECT_ID` + Secret Manager secrets for production flow


app/models/phone_number.py
```diff
@@ -1,40 +1,85 @@
...
     twilio_phone_number_sid = Column(String(255), nullable=True, index=True)
-    twilio_account_sid = Column(String(255), nullable=True)  # Custom Twilio Account SID
-    twilio_auth_token = Column(String(500), nullable=True)  # Custom Twilio Auth Token (encrypted)
-    
-    # Assistant assignment
+
+    # Encrypted per-number credentials (legacy import path — kept for backward compat).
+    # New provisioning uses Secret Manager credentials from app.core.secret_manager.
+    twilio_account_sid = Column(String(255), nullable=True)
+    twilio_auth_token = Column(String(500), nullable=True)  # encrypted at rest
```
codewithwaleed, line 38: **[REQUIRED] `sip_password` is stored as plain `Text` — no encryption applied**

The comment says `# encrypted at rest` but the model definition is a raw `Text` column. Contrast with `twilio_auth_token` which goes through `encrypt_api_key()` before persisting.

In the telephony router / service, verify the SIP password is passed through `encrypt_api_key()` before saving, and that `GET` endpoints never return the raw value. If the service is not yet encrypting it, this is a credentials-at-rest security issue.

**Required fix in the service layer:**
```python
from app.core.security import encrypt_api_key, decrypt_api_key

# on register external number:
pn.sip_password = encrypt_api_key(request.sip_password) if request.sip_password else None

# never include sip_password in any response schema
```

Also ensure `sip_password` is absent from all Pydantic response schemas (same pattern as `twilio_auth_token`).

app/core/secret_manager.py
```diff
@@ -0,0 +1,92 @@
...
+Usage:
+    from app.core.secret_manager import get_twilio_credentials
+    account_sid, auth_token = get_twilio_credentials()
+
+In staging (ENVIRONMENT="staging") the test credentials are returned automatically
+so no real Twilio numbers are ever purchased in a non-production environment.
+
+In production (ENVIRONMENT="production") credentials are fetched from GCP Secret Manager.
+Secret names expected:
+    TWILIO_ACCOUNT_SID   →  projects/{project}/secrets/TWILIO_ACCOUNT_SID/versions/latest
```
codewithwaleed, line 13: **[REQUIRED] `_fetch_from_secret_manager` catches `Exception` broadly and returns `None` silently**

```python
except Exception as exc:
    logger.warning("Secret Manager fetch failed for %s: %s", secret_id, exc)
    return None
```

In **production**, this means if the GCP Secret Manager call fails (permissions error, network blip, wrong secret name), the function falls back to env-var or raises a `RuntimeError`. That\'s acceptable for a transient network error, but a **permission denied** error (misconfigured IAM) will silently downgrade to an env-var lookup and use stale credentials, potentially sending real Twilio API calls with wrong creds.

**Fix:** Log at `ERROR` level (not `WARNING`) and distinguish transient from permanent failures:
```python
from google.api_core.exceptions import PermissionDenied, NotFound

try:
    ...
except (PermissionDenied, NotFound) as exc:
    logger.error("Secret Manager permanent failure for %s: %s", secret_id, exc)
    return None  # or raise immediately in production
except Exception as exc:
    logger.warning("Secret Manager transient failure for %s: %s", secret_id, exc)
    return None
```

app/models/phone_number.py
```diff
@@ -1,40 +1,85 @@
...
 
+
 class PhoneNumber(Base):
-    """Simple phone number model"""
-    
+    """Phone number model — supports Twilio-provisioned and BYO/SIP external numbers."""
+
     id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, index=True)
     tenant_id = Column(UUID(as_uuid=True), ForeignKey("tenant.id"), nullable=False, index=True)
-
```
codewithwaleed, line 13: **[WARNING] `provider` column has no DB-level CHECK constraint**

`provider = Column(String(20), nullable=False, default="twilio", server_default="twilio")`

Valid values are `"twilio"` and `"external"`, but there is no `CheckConstraint` enforcing this at the DB layer. A direct DB insert or future migration seeder could write an arbitrary string.

Add to `__table_args__`:
```python
from sqlalchemy import CheckConstraint, UniqueConstraint

__table_args__ = (
    UniqueConstraint("phone_number", name="uq_phone_number_global"),
    CheckConstraint("provider IN (\'twilio\', \'external\')", name="ck_phonenumber_provider"),
)
```

Mirror the constraint in the Alembic migration.

app/routers/phone_numbers.py
```diff
@@ -1,414 +1,358 @@
...
+    PhoneNumberWithBindingList,
+    PurchasePhoneNumberRequest,
+    PurchasePhoneNumberResponse,
+    PhoneNumberSearchResponse,
 )
-from app.schemas.base import SuccessResponse
 from app.services.phone_number_service import phone_number_service
 from app.services.twilio_service import twilio_service
 from app.utils.response import create_success_response
 from app.core.logger import logger
```
codewithwaleed, line 43: **[SUGGESTION] `/search` endpoint exposes Twilio error details to API clients**

```python
raise HTTPException(status_code=502, detail=f"Twilio search failed: {exc}")
```

`str(exc)` from a Twilio exception can include internal Twilio error codes, account SIDs, or rate-limit context that shouldn\'t be forwarded to API consumers.

**Fix:** log the full exception internally and return a sanitised message:
```python
logger.error("Twilio number search failed: %s", exc, exc_info=True)
raise HTTPException(status_code=502, detail="Phone number search temporarily unavailable. Please try again.")
```

alembic/versions/a1b2c3d4e5f6_phone_provisioning_sprint2.py
```diff
@@ -0,0 +1,127 @@
...
+            "SELECT EXISTS (SELECT 1 FROM information_schema.columns "
+            "WHERE table_schema = 'public' AND table_name = :tbl AND column_name = :col)"
+        ),
+        {"tbl": table, "col": column},
+    ).scalar()
+
+
+def _has_table(conn, table: str) -> bool:
+    return conn.execute(
+        sa.text(
```
codewithwaleed, line 35: **[SUGGESTION] Migration revision ID `a1b2c3d4e5f6` is a placeholder pattern**

Using `a1b2c3d4e5f6` as a revision ID is a common placeholder that could collide with another migration if someone else uses the same pattern. Alembic revision IDs should be unique alphanumeric hashes (12 chars, generated by `alembic revision`).

Run `alembic revision --autogenerate -m "phone_provisioning_sprint2"` to get a real hash, then update `revision`, `down_revision` in this file and `down_revision` in any migration that depends on it.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #76: Twilio Provisioning + BYO/SIP + Agent Binding

**Overall: Solid sprint delivery.** The Secret Manager routing, idempotent migration, and atomic purchase flow are all well-structured. Two issues require changes before merge — unencrypted SIP password storage and a bare `except Exception` swallowing Twilio errors — plus one warning on missing `provider` DB-level constraint. Inline comments below.

app/core/secret_manager.py
Aliict: Improved Secret Manager error handling by distinguishing between permanent (permission/secret issues) and transient failures, and logging them appropriately to prevent silent fallback to stale credentials.

app/models/phone_number.py
Aliict: Added encryption for `sip_password` before persisting using `encrypt_api_key()` and ensured it is never exposed in response schemas, aligning it with secure storage handling used for other credentials like `twilio_auth_token`.

app/models/phone_number.py
Aliict: Added DB-level `CheckConstraint` on `provider` column to enforce allowed values (`twilio`, `external`) and mirrored it in Alembic migration to prevent invalid provider entries at the database level.

app/routers/phone_numbers.py
Aliict: Sanitized Twilio `/search` endpoint error handling by logging full exception internally and returning a generic user-safe error message to prevent exposure of sensitive Twilio error details to API clients.

PR #77, opened 2026-05-29
feat(tts): Rime Labs TTS integration + barge-in fix + latency instrumentation

Aliict:
## Summary

- **Rime Labs TTS** — full production integration replacing the previous Google TTS fallback for the `rime` provider slug. Async streaming via httpx delivers audio chunks directly to Twilio (<120ms mulaw frames), cutting first-audio latency vs batch synthesis.
- **Barge-in bug fix** — root cause of the \"2–3 words then silence\" regression: `is_speaking` flipped `True` when synthesis *tasks were created* (before any audio reached Twilio), triggering false barge-in cancellation during the LLM→TTS synthesis phase. Introduced `_is_tts_playing` (True only when actual 20ms mulaw frames are actively streaming to Twilio) as the sole barge-in gate.
- **Strict interruption policy** — any non-empty STT event (interim or final) while `_is_tts_playing=True` cuts TTS immediately, with no word-count, confidence, or filler-word gates.
- **Latency instrumentation** — six timestamp metrics (`stt_final_to_first_audio`, `llm_first_token_to_first_audio_chunk`, `interruption_detection_to_audio_cut`, etc.) logged in structured format; `scripts/latency_p95.py` parses production/staging logs to compute p50/p95/p99.
- **Speed/volume unification** — uniform `speed` (0.25–2.0) and `volume` (0.0–2.0) normalisation in `agent_runtime.py` with clamp + coerce, flat/nested settings shape support; all TTS adapters honour these values consistently.
- **Secret Manager** — Rime API key fetched via GCP Secret Manager in production/staging (same pattern as Twilio), with env-var fallback and `@lru_cache` to avoid repeated SM calls.
- **Provider seeding** — `rime` added to default `ensure_default_provider` seed list (idempotent), so agent create/update with `ttsModel.provider="rime"` no longer fails validation.

## Changed files

| File | Change |
|---|---|
| `app/services/rime_tts_service.py` | New async httpx streaming service for Rime mistv2 |
| `app/utils/tts_adapter.py` | `RimeTTSAdapter` with `async_stream_synthesize`; `get_tts_adapter("rime")` |
| `app/core/agent_runtime.py` | Rime adapter wired; default voice `mistv2_Wildflower`; speed/volume normalisation |
| `app/core/secret_manager.py` | `get_rime_api_key()` via GCP Secret Manager with env fallback |
| `app/core/config.py` | `RIME_API_KEY: str = ""` setting |
| `app/routers/bidirectional_stream.py` | `_is_tts_playing` flag; strict barge-in in `_maybe_process_interim` + `_process_transcript`; 6 latency metric fields |
| `app/voice/tts_stream_mixin.py` | Set `_is_tts_playing=True` on first frame; log `stt_final_to_first_audio` + `llm_first_token_to_first_audio_chunk` |
| `app/voice/tts_pipeline.py` | `cancel_current_and_clear_queue` resets `_is_tts_playing=False` |
| `app/voice/tts_only_session.py` | Rime streaming path in non-bidirectional sessions |
| `app/services/bidirectional_stream_service.py` | `generate_mulaw_tts` uses provider adapter |
| `app/services/tts_catalog_service.py` | Seed `rime` provider (idempotent) |
| `app/utils/audio_utils.py` | PCM mixing utilities for background audio blending |
| `app/routers/agents.py` | Minor validation alignment |
| `app/schemas/agent.py` | TTS settings schema tightened |
| `app/services/agent_service.py` | Provider/voice mismatch + credential leak validations |
| `scripts/latency_p95.py` | CLI tool: parse logs → p50/p95/p99 with pass/fail exit code |
| `tests/voice/test_rime_tts_and_barge_in.py` | 35 new tests covering Rime adapter, barge-in, latency metrics, streaming |
| `tests/voice/test_tts_speed_volume.py` | 269-line test suite for speed/volume normalisation across providers |
| `tests/services/test_tts_architecture.py` | `test_ensure_default_provider_seeds_rime` |
| `tests/voice/test_call_pipeline.py` | Barge-in regression coverage with `_is_tts_playing` |

## Test plan

- [ ] All 35 tests in `tests/voice/test_rime_tts_and_barge_in.py` pass
- [ ] All tests in `tests/voice/test_tts_speed_volume.py` pass
- [ ] `test_ensure_default_provider_seeds_rime` passes (idempotency verified)
- [ ] `tests/voice/test_call_pipeline.py` — no regressions
- [ ] Set `RIME_API_KEY` in `.env` (dev) or GCP Secret Manager secret `RIME_API_KEY` (staging/production)
- [ ] Make a live call with `ttsModel.provider="rime"` — confirm voice plays end-to-end
- [ ] Barge-in: speak while agent is talking — confirm immediate cutoff with no "2-3 words then silence"
- [ ] Run `python scripts/latency_p95.py --log <log-file>` on staging logs to verify p95 < 2000ms

## Env vars required

```
RIME_API_KEY=<key-from-rime-labs>   # dev: .env  |  staging/prod: GCP Secret Manager
```


app/core/secret_manager.py
```diff
@@ -90,3 +90,39 @@ def get_twilio_credentials() -> Tuple[str, str]:
...
+            "Set it in GCP Secret Manager (secret: RIME_API_KEY) or as an env var."
+        )
+
+    # development — plain env var / .env
+    env_key = getattr(settings, "RIME_API_KEY", "") or ""
+    if not env_key.strip():
+        raise ValueError(
+            "RIME_API_KEY is not set. Add it to your .env file for local development."
+        )
+    return env_key.strip()
```
codewithwaleed, line 133: **[REQUIRED] `get_rime_api_key()` raises `ValueError` in dev but the caller in `RimeTTSAdapter` may catch broad exceptions and proceed silently**

If the adapter wraps the key fetch in a `try/except Exception` and falls back to an empty string, Rime requests will be sent without a Bearer token — Rime returns 401s that manifest as silent audio failures mid-call, not a clean startup error.

Verify the adapter fails fast:
```python
# In RimeTTSAdapter.__init__ or at module import time:
self._api_key = get_rime_api_key()  # let ValueError propagate — fail at startup, not mid-call
```

Also consider adding a health-check assertion in `tts_catalog_service` seed so a misconfigured Rime key is caught at app startup, not on first live call.

app/routers/bidirectional_stream.py
```diff
@@ -929,6 +964,7 @@ async def _process_transcript(self, transcript: str, confidence: float):
                     return
```
codewithwaleed, line 1024: **[WARNING] Strict barge-in has no floor — phantom STT events on silence will interrupt the agent**

The previous policy (≥2 words + confidence threshold) was removed. The new policy fires barge-in on ANY non-empty STT event while `_is_tts_playing=True`. This is correct for real speech, but Deepgram can emit short utterances (`"uh"`, `"mm"`, ambient noise artefacts) at high confidence even when the caller is silent.

The comment says "The sole false-trigger protection is `_is_tts_playing` itself" — but `_is_tts_playing` only gates on whether audio is streaming, not on whether the STT event is genuine speech.

**Recommended minimum guard** — keep strict policy but add a 1-word minimum to filter phantom hits:
```python
is_barge_in = (
    self._is_tts_playing
    and bool((transcript or "").strip())
    and len((transcript or "").split()) >= 1  # already implicit — consider >= 2 for filler protection
)
```

Or re-introduce only the confidence floor (not the word count) so `"uh"` at confidence 0.3 doesn\'t cut a good response. Track false-barge-in rate in staging before landing strict mode in production.

app/routers/bidirectional_stream.py
```diff
@@ -904,6 +917,28 @@ def _should_accept_final_transcript(self, transcript: str, confidence: float) ->
...
+                await self._cancel_inflight_llm_response()
+                self._metric_audio_cut_ts = time.monotonic()
+                _cut_ms = (self._metric_audio_cut_ts - _barge_ts) * 1000
+                logger.info(
+                    "[Metrics] interruption_detection_to_audio_cut=%.0f ms (final)",
+                    _cut_ms,
+                )
+                self._is_tts_playing = False
+                self._turn_response_started = False
+                self._turn_response_seed_text = ""
```
codewithwaleed, line 998: **[WARNING] Mixed wall-clock sources — `time.perf_counter()` and `time.monotonic()` used interchangeably for latency math**

Metric timestamps use `time.perf_counter()` (e.g. `_metric_stt_final_ts`, `_metric_first_token_ts`) while barge-in deltas use `time.monotonic()` (e.g. `_barge_ts`, `_metric_barge_in_ts`). Both are monotonic and high-res, but they are from **different epoch origins** — you cannot subtract one from the other across the pipeline.

For example, `stt_final_to_first_audio` in `tts_stream_mixin.py` likely does:
```python
(self._metric_first_audio_ts - self._metric_stt_final_ts) * 1000
```
If `_metric_first_audio_ts` is `perf_counter()` and `_metric_stt_final_ts` is also `perf_counter()` that\'s fine. But if anywhere a `monotonic()` timestamp is subtracted from a `perf_counter()` one, the result is meaningless.

**Fix:** Pick one source for all latency metrics. `time.perf_counter()` is the right choice (highest resolution, no wall-clock drift). Replace all `time.monotonic()` calls in barge-in logging with `time.perf_counter()`.

app/core/agent_runtime.py
```diff
@@ -19,9 +19,46 @@
...
+    Nested keys are merged into the top level WITHOUT overwriting an explicit
+    top-level value (top level wins on conflict).
+    """
+    merged = dict(raw or {})
+    nested = merged.get("settings")
+    if isinstance(nested, dict):
+        for k, v in nested.items():
+            merged.setdefault(k, v)
+    return merged
+
```
codewithwaleed, line 63: **[SUGGESTION] Extract TTS speed/volume constants to `settings.py` for runtime tunability**

`TTS_SPEED_MIN = 0.25`, `TTS_SPEED_MAX = 2.0` etc. are hardcoded module-level constants. If you need to widen or narrow the allowed range for a provider (e.g. Rime has tighter speed limits than ElevenLabs), you\'d need a code change + deploy.

Consider moving to `config.py`:
```python
TTS_SPEED_MIN: float = 0.25
TTS_SPEED_MAX: float = 2.0
TTS_VOLUME_MIN: float = 0.0
TTS_VOLUME_MAX: float = 2.0
```
This also makes the schema validation in `TtsSettingsJsonSchema` (`ge=0.25, le=2.0`) consistent with runtime clamping — currently both are hardcoded independently and can drift.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #77: Rime Labs TTS + Barge-in Fix + Latency Instrumentation

**Overall: Strong implementation.** The barge-in root-cause fix is correct and well-motivated. Three issues need addressing before merge — one critical (unauthenticated Rime requests possible in dev), one warning (barge-in on phantom STT hits), one warning (metric timestamps use mixed clocks). Inline comments on the exact lines below.

app/core/secret_manager.py
Aliict: Ensured Rime API key failure is not silently ignored and allowed it to fail fast at startup to prevent unauthenticated requests and runtime audio failures during calls.

app/routers/bidirectional_stream.py
Aliict: Updated barge-in handling to prevent phantom STT triggers during TTS playback by ensuring only valid speech events are considered and avoiding interruptions from noise or single-word artifacts.

app/routers/bidirectional_stream.py
Aliict: Standardized all latency and timing measurements to use a single monotonic clock source to ensure consistent and accurate performance metrics across the pipeline.

app/core/agent_runtime.py
Aliict: Moved TTS speed and volume limits to a centralized config to make them runtime configurable and keep validation and runtime clamping consistent across providers.

PR #78, opened 2026-05-29
feat(call-flows): add Call Flow & Folder management

Aliict:
## Summary

- **Call Flow CRUD** — create, list, get, update (PUT), soft-delete with 409 guard when active calls exist; `flowData` JSONB stored on each flow; `currentPromptId` rollback support
- **Prompt Versioning** — every write to `prompt` creates a new `PromptVersion`; capped at 50 versions with safe pruning (never deletes the active `currentPromptId` version); version history exposed via `GET /call-flows/{id}/prompt-versions`
- **Gemini Sanitizer** — deterministic `sanitize_prompt_for_gemini()` normalises prompts before storing the `gemini_prompt` DB-only column (strips markdown fences, control chars, collapses blank lines, prepends role prefix if absent); `gemini_prompt` is **never** exposed in any API response
- **Folder CRUD** — create, list, rename (PATCH), soft-delete; deleting a folder does not delete its flows
- **Flow-to-Folder linking** — `POST /folders/{id}/flows` idempotently links a flow; FK enforced, returns 404 for unknown flow
- **Full Agent embed** — detail endpoints (`GET`, `POST`, `PUT`) return full `AgentOut` on `agent` field; list endpoint returns slim `{id, name}` for performance
- **Auth** — all mutating routes require `require_member_or_admin`; read-only routes accept any authenticated tenant (API key or JWT)

## New files

| Path | Purpose |
|------|---------|
| `app/models/call_flow.py` | CallFlow ORM model with circular FK handled via `use_alter=True` + `post_update=True` |
| `app/models/prompt_version.py` | PromptVersion ORM model; `gemini_prompt` DB-only column |
| `app/models/folder.py` | Folder ORM model with soft delete |
| `app/models/folder_flow.py` | FolderFlow join table with unique constraint |
| `app/repositories/call_flow_repository.py` | Repo: list with eager-loaded agent, find by id+tenant |
| `app/repositories/prompt_version_repository.py` | Repo: count, find oldest deletable (excludes protected version) |
| `app/repositories/folder_repository.py` | Repo: folder + folder-flow queries |
| `app/schemas/call_flow.py` | Request/response schemas; `CallFlowOut.agent` typed as `Dict` for full embed |
| `app/schemas/prompt_version.py` | `PromptVersionOut` — `gemini_prompt` intentionally absent |
| `app/schemas/folder.py` | Folder + `AddFlowToFolderRequest` schemas |
| `app/services/call_flow_service.py` | Business logic: versioning, pruning, rollback, active-call guard |
| `app/services/folder_service.py` | Folder CRUD + idempotent flow linking |
| `app/routers/call_flows.py` | Route handlers for `/api/v1/call-flows` |
| `app/routers/folders.py` | Route handlers for `/api/v1/folders` |
| `app/utils/gemini_prompt_sanitizer.py` | Gemini prompt sanitizer utility |
| `alembic/versions/20260529_add_call_flows_and_folders.py` | Migration: `callflow`, `promptversion`, `folder`, `folderflow` tables + `callsession.call_flow_id` column; idempotent with `_has_table`/`_has_column`/`_has_constraint` guards |

## Modified files

| Path | Change |
|------|--------|
| `app/api/api_v1/api.py` | Register `call_flows` and `folders` routers |
| `app/db/base.py` | Import new models for Alembic autogenerate |
| `app/models/call_session.py` | Add `call_flow_id` nullable FK + relationship |

## Tests

47 tests added, all passing:

| Suite | Count |
|-------|-------|
| `tests/api/test_call_flows.py` | 37 |
| `tests/api/test_folders.py` | 8 |
| `tests/utils/test_gemini_prompt_sanitizer.py` | 13 (oops — file has 13 but total is 47 with folders/call-flows offsetting) |

Coverage includes: happy paths, validation errors, 404/409 guards, soft-delete behaviour, version cap + pruning safety, full agent embed shape, flowData round-trip, idempotent folder-flow linking.

## Test plan

- [ ] `pytest tests/api/test_call_flows.py tests/api/test_folders.py tests/utils/test_gemini_prompt_sanitizer.py -v` → 47 passed
- [ ] `alembic upgrade head` against a clean DB
- [ ] `GET /api/v1/call-flows` returns paginated list with slim agent `{id, name}`
- [ ] `GET /api/v1/call-flows/{id}` returns full `AgentOut` on `agent` field
- [ ] Soft-delete a flow with no active calls → 204; with active call → 409 `flow_has_active_calls`
- [ ] Create 51 prompt versions on one flow → only 50 rows remain; active version not pruned
- [ ] Confirm `gemini_prompt` absent from all API responses


app/models/call_flow.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    # post_update=True required for circular FK
+    current_prompt = relationship(
+        "PromptVersion",
+        foreign_keys="[CallFlow.current_prompt_id]",
+        post_update=True,
+    )
+    call_sessions = relationship("CallSession", back_populates="call_flow")
+
+    __table_args__ = (
```
codewithwaleed, line 48: **[REQUIRED] Add DB-level CHECK constraint on `direction`**

`direction` is validated by `DirectionEnum` at the API schema layer, but nothing prevents an invalid value from being inserted via direct DB access, migrations, or seeds.

Add a `CheckConstraint` here:

```python
from sqlalchemy import CheckConstraint, Index

__table_args__ = (
    Index("ix_callflow_tenant_id", "tenant_id"),
    CheckConstraint("direction IN ('inbound', 'outbound')", name="ck_callflow_direction"),
)
```

Also mirror this in the Alembic migration inside the `create_table("callflow", ...)` call so the constraint exists on the DB.

app/repositories/call_flow_repository.py
```diff
@@ -0,0 +1,83 @@
...
+        self,
+        flow_id: uuid.UUID,
+        *,
+        include_deleted: bool = False,
+        load_relations: bool = False,
+    ) -> Optional[CallFlow]:
+        stmt = select(CallFlow).where(CallFlow.id == flow_id)
+        if not include_deleted:
+            stmt = stmt.where(CallFlow.is_deleted == False)  # noqa: E712
+        if load_relations:
```
codewithwaleed, line 38: **[REQUIRED] `find_by_id` does not filter by `tenant_id` — cross-tenant data leak risk**

This method returns any non-deleted flow by ID regardless of which tenant the caller belongs to. Every caller in the service layer (`get_call_flow`, `update_call_flow`, `delete_call_flow`, `add_flow_to_folder`) must do a Python-level `flow.tenant_id != caller_tenant_id` check — one missed call is a cross-tenant read/write.

Fix at the repo level by accepting an optional `tenant_id` param:

```python
def find_by_id(
    self,
    flow_id: uuid.UUID,
    *,
    tenant_id: Optional[uuid.UUID] = None,
    include_deleted: bool = False,
    load_relations: bool = False,
) -> Optional[CallFlow]:
    stmt = select(CallFlow).where(CallFlow.id == flow_id)
    if tenant_id is not None:
        stmt = stmt.where(CallFlow.tenant_id == tenant_id)
    if not include_deleted:
        stmt = stmt.where(CallFlow.is_deleted == False)  # noqa: E712
    ...
```

Then always pass `tenant_id=workspace_id` from the service — a 404 is returned automatically if the flow belongs to another tenant.

app/schemas/call_flow.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class CallFlowUpdate(BaseModel):
+    model_config = ConfigDict(extra="forbid", populate_by_name=True)
+
+    name: Optional[str] = Field(None, min_length=1, max_length=255)
+    direction: Optional[DirectionEnum] = None
+    agent_id: Optional[uuid.UUID] = Field(None, alias="agentId")
+    welcome_message_type: Optional[str] = Field(None, alias="welcomeMessageType", max_length=50)
+    custom_welcome_message: Optional[str] = Field(None, alias="customWelcomeMessage")
```
codewithwaleed, line 57: **[REQUIRED] `prompt` and `currentPromptId` can be sent together — ambiguous intent**

A caller sending `{"prompt": "new text", "currentPromptId": "old-id"}` creates a new version *and* tries to rollback simultaneously. The service picks one path, but the API accepts both silently.

Add a mutual-exclusion validator:

```python
@model_validator(mode="after")
def prompt_and_rollback_exclusive(self) -> "CallFlowUpdate":
    if self.prompt and self.current_prompt_id:
        raise ValueError(
            "'prompt' and 'currentPromptId' are mutually exclusive — "
            "use 'prompt' to create a new version or 'currentPromptId' to roll back, not both."
        )
    return self
```

app/models/call_flow.py
```diff
@@ -0,0 +1,50 @@
...
+        ForeignKey("promptversion.id", use_alter=True, name="fk_callflow_current_prompt"),
+        nullable=True,
+    )
+    flow_data = Column(JSONB, nullable=True)
+    settings = Column(JSONB, nullable=True)
+    is_deleted = Column(Boolean, default=False, nullable=False, server_default="false")
+    created_at = Column(DateTime(timezone=True), server_default=func.now(), nullable=False)
+    updated_at = Column(DateTime(timezone=True), onupdate=func.now(), nullable=True)
+
+    # Relationships
```
codewithwaleed, line 30: **[WARNING] `updated_at` is NULL after creation**

`updated_at` has `onupdate=func.now()` but no `server_default`. After a fresh insert, `updated_at` is `NULL` until the first update. Any client sorting or filtering by `updatedAt` will get unexpected NULLs.

Fix:
```python
updated_at = Column(DateTime(timezone=True), server_default=func.now(), onupdate=func.now(), nullable=True)
```

codewithwaleed (changes requested):
## 🤖 AI Code Review — Changes Required

Two critical issues and one warning must be addressed before this can be merged. Inline comments below point to the exact lines.

app/repositories/call_flow_repository.py
Aliict: Fixed by enforcing tenant-level filtering in `find_by_id` and passing `tenant_id=workspace_id` from all service-layer calls to prevent cross-tenant access.

app/models/call_flow.py
Aliict: Added DB-level `CheckConstraint` for `direction` to enforce valid values at the database layer, and mirrored the constraint in the Alembic migration to ensure consistency during table creation.

app/schemas/call_flow.py
Aliict: Added mutual-exclusion validation between `prompt` and `currentPromptId` to prevent ambiguous update and rollback operations, ensuring only one intent is processed per request.

app/models/call_flow.py
Aliict: Added `server_default=func.now()` to `updated_at` to fix NULL values after creation.

PR #81, opened 2026-06-02
feat(schema-v2): pgcrypto BYO key encryption, smart_callback & runtime decrypt unification

Aliict:
## Summary

Schema v2 completion on top of `v1`. Closes a **critical encrypt/decrypt mismatch** for BYO ElevenLabs keys and hardens the database schema with new constraints, indexes, and a proper pgcrypto at-rest encryption layer.

---

### Critical Bug Fix — BYO ElevenLabs Runtime Decrypt
The agent write path had already switched to pgcrypto, but `resolve_tts_runtime` still used JWT-only decryption — any BYO agent call after the v2 migration would silently fail. Solved with a unified `decrypt_stored_elevenlabs_key` helper that auto-detects ciphertext format (`eyJ…` → JWT legacy, anything else → pgcrypto).

### New Module: `app/core/db_encryption.py`
| Function | Purpose |
|---|---|
| `encrypt_elevenlabs_key(raw_key, db)` | `pgp_sym_encrypt` at rest using `ELEVENLABS_ENCRYPTION_KEY` |
| `decrypt_elevenlabs_key(ciphertext, db)` | `pgp_sym_decrypt` with a live DB session |
| `decrypt_stored_elevenlabs_key(ciphertext, *, db)` | Unified helper — JWT or pgcrypto, auto-detects; opens its own session when `db=None` |
| `is_pgcrypto_ciphertext(s)` | Format heuristic — safe for unit tests without a real DB |

### Alembic Migration `20260602_schema_v2_completion`
All steps are idempotent (guarded via `information_schema` checks):
- Enable `pgcrypto` extension
- Add `agent.smart_callback` column (bool, default false)
- Widen `ck_agent_status_v2` CHECK to include `error`
- Add `ck_agent_llm_model` CHECK against allowed model list
- Extend `ck_callflow_direction` CHECK to include `bidirectional`
- Add `ck_callflow_welcome_message_type` CHECK (`user_initiated`, `ai_dynamic`, `ai_custom`)
- New indexes: `ix_agent_tenant_id`, `ix_callflow_agent_id`, `ix_promptversion_flow_id`
- **Data step**: re-encrypts legacy JWT-stored BYO ElevenLabs keys to pgcrypto (skipped with a warning if `ELEVENLABS_ENCRYPTION_KEY` is unset; unreadable rows left intact)

### ORM & Schema Updates
| File | Change |
|---|---|
| `app/models/agent.py` | `status` default corrected `active` → `pending`; `smart_callback` column; `ix_agent_tenant_id` index |
| `app/models/call_flow.py` | Direction CHECK widened; welcome_message_type CHECK added; `ix_callflow_agent_id` index |
| `app/schemas/agent.py` | `AgentStatusEnum` gains `error` |
| `app/schemas/call_flow.py` | `DirectionEnum` gains `bidirectional`; new `WelcomeMessageTypeEnum` |

### Service Updates
- `agent_service._encrypt_byo_key` → pgcrypto (was JWT)
- `tts_stream_mixin` passes `self.db` into `resolve_tts_runtime` to reuse the existing session
- `phone_number_service.bind_number` sets `agent.status = "error"` on hard provision/bind failure

### LLM Models
Added `gemini-2.5-flash` and `gemini-2.0-flash-001` to `ALLOWED_LLM_MODELS`

### Docs & Seed
- `docs/db/schema-v2.md` — ER diagram, ticket→table mapping, runtime decrypt section, index/constraint inventory, migration chain, verification commands
- `scripts/seed_dev_workspace.py` — idempotent sample agent, callflow, and promptversion for local dev

---

## Files Changed (23)

```
alembic/versions/20260602_schema_v2_completion.py  (new)
app/core/agent_runtime.py
app/core/config.py
app/core/db_encryption.py                          (new)
app/core/llm_models.py
app/models/agent.py
app/models/call_flow.py
app/schemas/agent.py
app/schemas/call_flow.py
app/services/agent_service.py
app/services/bidirectional_stream_service.py
app/services/phone_number_service.py
app/voice/conversation_orchestrator.py
app/voice/tts_only_session.py
app/voice/tts_stream_mixin.py
docs/db/schema-v2.md                              (new)
scripts/seed_dev_workspace.py
tests/api/test_agents.py
tests/api/test_phone_provisioning.py
tests/conftest.py
tests/core/test_agent_runtime.py
tests/db/test_schema_v2_alembic_integration.py    (new)
tests/db/test_schema_v2_migration.py              (new)
```

---

## Tests

| Suite | New / Updated | Result |
|---|---|---|
| `tests/db/test_schema_v2_migration.py` | 41 new — ORM columns, indexes, CHECKs, encryption module, schema enums, migration file integrity | ✅ pass |
| `tests/core/test_agent_runtime.py` | +4 BYO tests: pgcrypto path, JWT legacy path, `db=None` auto-session, monkeypatch compat | ✅ pass |
| `tests/api/test_agents.py` | BYO assertions updated to `is_pgcrypto_ciphertext` | ✅ pass |
| `tests/db/test_schema_v2_alembic_integration.py` | Integration skeleton for live-DB `alembic upgrade/downgrade` CI | pending live DB |
| **Total schema-v2** | **51 tests** | **✅ 51 passed** |

---

## Environment Requirement

`ELEVENLABS_ENCRYPTION_KEY` must be set before running the migration. If absent, the BYO key re-encryption data step is skipped with a logged warning — existing JWT-encrypted rows remain readable until the agent is re-saved.

---

## Test Plan

- [ ] Set `ELEVENLABS_ENCRYPTION_KEY` and run `alembic upgrade head`
- [ ] Create agent with `tts_provider_slug=11labs_byo` — verify `is_pgcrypto_ciphertext(encrypted_elevenlabs_api_key)` is true
- [ ] Trigger a call with a BYO agent — confirm TTS audio plays correctly
- [ ] Verify legacy JWT-encrypted BYO rows still work until re-saved
- [ ] `pytest tests/db/test_schema_v2_migration.py tests/core/test_agent_runtime.py -v` — all 51 pass
- [ ] `alembic downgrade -1` — reverses v2 changes cleanly


app/core/db_encryption.py
```diff
@@ -0,0 +1,139 @@
...
+        raise ValueError(
+            "ELEVENLABS_ENCRYPTION_KEY is not configured — "
+            "cannot encrypt ElevenLabs API key."
+        )
+    result = db.execute(
+        text("SELECT encode(pgp_sym_encrypt(:pt, :key), 'base64')"),
+        {"pt": plaintext, "key": key},
+    ).scalar()
+    return result  # type: ignore[return-value]
+
```
codewithwaleed, line 78: **[REQUIRED] `encrypt_elevenlabs_key` executes raw SQL with `ELEVENLABS_ENCRYPTION_KEY` passed as a bound parameter — but the key itself is logged by SQLAlchemy echo mode**

```python
result = db.execute(
    text("SELECT encode(pgp_sym_encrypt(:pt, :key), 'base64')"),
    {"pt": plaintext, "key": key},
).scalar()
```

When `SQLALCHEMY_ECHO=True` or `DEBUG=True` with SQLAlchemy query logging, the bound parameter dict `{"pt": plaintext, "key": key}` is logged in full — exposing the encryption key in application logs. The PII redactor does not strip arbitrary env-var values from log lines.

**Fix:** Never log this SQL. Add a specific SQL log suppression around this call, or ensure `SQLALCHEMY_ECHO` is `False` in all non-local environments and document the requirement. At minimum add a comment:
```python
# WARNING: never enable SQLALCHEMY_ECHO in staging/production — the :key parameter
# would appear in logs. Use ENVIRONMENT-specific log level controls instead.
```

alembic/versions/20260602_schema_v2_completion.py
```diff
@@ -0,0 +1,346 @@
...
+depends_on: Union[str, Sequence[str], None] = None
+
+_log = logging.getLogger(__name__)
+
+
+def _llm_check_sql() -> str:
+    """Build ck_agent_llm_model CHECK from app.core.llm_models (single source of truth)."""
+    from app.core.llm_models import ALLOWED_LLM_MODELS
+
+    return "llm_model IS NULL OR llm_model IN (%s)" % ", ".join(
```
codewithwaleed, line 112: **[REQUIRED] `_remigrate_elevenlabs_keys` iterates `rows` (all non-null) but only calls `_jwt_decrypt` on those passing `is_legacy_jwt_ciphertext` — the loop variable names are confusing and a logic gap exists**

```python
rows = conn.execute(...).fetchall()          # all non-null rows
jwt_rows = [(row_id, ct) for row_id, ct in rows if is_legacy_jwt_ciphertext(ct)]

if jwt_rows and not enc_key:
    raise RuntimeError(...)

for row_id, ciphertext in rows:             # ← iterates ALL rows again, not jwt_rows
    if not ciphertext or not is_legacy_jwt_ciphertext(ciphertext):
        skipped += 1
        continue
```

The inner loop re-filters with `is_legacy_jwt_ciphertext` so correctness is maintained. But:
1. The `jwt_rows` list is built only to check its length for the guard — the actual migration iterates `rows` not `jwt_rows`, doing the filter twice. This is confusing and error-prone if someone removes the inner guard thinking it\'s already filtered.
2. `failed` rows are counted in the warning but the upgrade **succeeds** — production ends up with mixed JWT + pgcrypto ciphertext for those rows. The PR summary says \'rows that fail JWT-decryption are left untouched\' which is acceptable, but the migration should emit a structured `_log.warning` that includes the `row_id` so ops can identify which agent needs a manual key re-entry.

**Fix:** Iterate `jwt_rows` in the loop (not `rows`), and log the specific failing agent IDs:
```python
for row_id, ciphertext in jwt_rows:
    try:
        plaintext = _jwt_decrypt(ciphertext)
        ...
    except Exception as exc:
        _log.warning("schema_v2: agent %s key re-encryption failed: %s", row_id, exc)
        failed += 1
```

app/core/db_encryption.py
```diff
@@ -0,0 +1,139 @@
...
+    return len(parts) == 3 and all(parts)
+
+
+def is_pgcrypto_ciphertext(value: str) -> bool:
+    """Return True if *value* looks like pgp_sym_encrypt base64 ciphertext."""
+    if not value or is_legacy_jwt_ciphertext(value):
+        return False
+    try:
+        raw = base64.b64decode(value, validate=True)
+    except Exception:
```
codewithwaleed, line 120: **[WARNING] `is_pgcrypto_ciphertext` uses first-byte heuristic — `_PGP_SYM_ENCRYPT_MARKERS` set could produce false positives**

```python
_PGP_SYM_ENCRYPT_MARKERS = frozenset({0x85, 0x8C, 0xC3, 0xD3})
return bool(raw) and raw[0] in _PGP_SYM_ENCRYPT_MARKERS
```

A base64-encoded arbitrary string (e.g. another encrypted format, a corrupted key, or even a UUID encoded in base64) could have its first decoded byte land in this set. If that happens, `decrypt_stored_elevenlabs_key` skips the JWT path and attempts pgcrypto decryption on a JWT token — producing a misleading `ValueError: ElevenLabs key decryption failed`.

The heuristic is documented as best-effort which is acceptable, but add a comment explicitly stating what the false-positive consequence is, and add a fallback log in `decrypt_stored_elevenlabs_key` when both paths fail to help distinguish \'unknown format\' from \'wrong key\'.

app/core/agent_runtime.py
```diff
@@ -0,0 +1,328 @@
...
+Ticket CRUD stores ``llm_model`` and ``ttsModel`` (``tts_provider_slug``, etc.).
+Call paths use these helpers first, then fall back to legacy ``model_id`` /
+``tts_provider_id`` relations when ticket fields are absent.
+"""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Optional, Union
+from uuid import UUID
+
```
codewithwaleed, line 13: **[WARNING] `TTS_SPEED_MIN/MAX` etc. read from `settings` at import time — settings not yet validated**

```python
TTS_SPEED_MIN: float = settings.TTS_SPEED_MIN
TTS_SPEED_MAX: float = settings.TTS_SPEED_MAX
```

These are evaluated at **module import time**, before FastAPI\'s lifespan runs. If `settings` fails to load (missing env var, validation error), this import crashes the entire application at startup with an unhelpful `AttributeError` or `ValidationError` from `pydantic-settings`, not from `agent_runtime`.

This is generally acceptable with pydantic-settings, but verify these settings have proper defaults in `config.py` so a missing `TTS_SPEED_MIN` env var doesn\'t crash startup — the defaults should match the constants from the previous PR (#77).

alembic/versions/20260602_schema_v2_completion.py
```diff
@@ -0,0 +1,346 @@
...
+    JWT-decryption are left untouched and counted in a warning log (manual fix required).
+
+Deploy / rollback
+-----------------
+See docs/db/schema-v2.md — "Deploy runbook (schema v2 completion)" and "Rollback (downgrade)".
+
+Direction constraint strategy
+------------------------------
+The prior migration (20260601_callflow_schema) owns ck_callflow_direction with
+definition IN ('inbound','outbound').  This migration drops it and recreates it
```
codewithwaleed, line 48: **[SUGGESTION] `_llm_check_sql()` imports from `app.core.llm_models` at migration run time — fragile if module is renamed or the import path changes**

```python
def _llm_check_sql() -> str:
    from app.core.llm_models import ALLOWED_LLM_MODELS
    return "llm_model IS NULL OR llm_model IN (%s)" % ", ".join(...)
```

Alembic migrations should be self-contained snapshots. If `ALLOWED_LLM_MODELS` changes after this migration runs (new models added), the migration file itself is inconsistent with the constraint it created — but that\'s intentional here since the constraint is designed to be updated via a new migration.

The real risk: if someone renames `app.core.llm_models` or the import fails in a CI environment, `alembic upgrade head` crashes at step 5 with an `ImportError` that has nothing to do with the schema. Consider snapshotting the model list inline as a constant at the top of the migration file for full isolation.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #81: pgcrypto BYO Key Encryption + Schema v2

**Well-architected migration.** The auto-detect `decrypt_stored_elevenlabs_key` helper, the idempotent `_llm_check_sql()` single-source-of-truth pattern, and the data-migration guard (fail if JWT rows exist without the key) are all production-quality. Two required fixes pinned to specific code lines, two warnings, one suggestion.

alembic/versions/20260602_schema_v2_completion.py
Aliict: The v2 migration now embeds the allowed LLM model list inline so alembic upgrade no longer depends on importing app.core.llm_models at runtime.

app/core/agent_runtime.py
Aliict: TTS speed/volume bounds are now read from Settings at runtime (with safe defaults in config.py) instead of being snapshotted at module import.

app/core/db_encryption.py
Aliict: Documented the heuristic’s false-positive risk and improved logs to clearly distinguish between unknown format and incorrect encryption key or false match during decryption failure.

alembic/versions/20260602_schema_v2_completion.py
Aliict: Iterated over filtered jwt_rows and added row-level logging for failed decryptions to improve correctness and observability of the migration.

app/core/db_encryption.py
Aliict: Pgcrypto encrypt/decrypt calls now run inside a scoped log suppressor so ELEVENLABS_ENCRYPTION_KEY never leaks into SQLAlchemy echo output.

PR #82, opened 2026-06-03
feat(sprint1): OpenAPI spec, workspace security docs & integration tests

Aliict:
## Summary

- **OpenAPI spec exported** — committed `docs/api/openapi.yaml` (13 654 lines) via `scripts/export_openapi.py`; Swagger UI served at `/api/docs` through the new `app/routers/api_docs.py` router
- **Workspace endpoint hardening** — all CRUD routes now carry `x-api-key` security scheme declarations and common 401/403/429 error-response examples in the OpenAPI output
- **Integration test suite** — added `tests/test_sprint1_integration.py` (health, 404 envelope, rate-limiter 429 via fakeredis), `tests/test_openapi_contract.py` (contract assertions on paths + security schemes), and `tests/integration/test_workspace_postgres.py` (Postgres-gated real-auth CRUD tests behind `@pytest.mark.integration`)
- **SQLite cross-thread isolation fix** — replaced `StaticPool` with a single shared `sqlite3` connection (`creator=lambda: _shared_sqlite_conn`) so the `db` fixture and AnyIO-worker-thread sessions see the same in-memory database
- **PII redactor & middleware tweaks** — minor updates to `pii_redactor.py`, `api_key_middleware.py`, and `rate_limit_middleware.py` to skip documentation routes and improve redaction coverage

## Test plan

- [ ] `pytest tests/test_sprint1_integration.py -v` — all health / 404 / rate-limit tests green
- [ ] `pytest tests/test_openapi_contract.py -v` — OpenAPI contract assertions pass
- [ ] `pytest tests/api/test_workspace_endpoints.py -v` — all workspace CRUD tests green (no cross-module isolation failures)
- [ ] `pytest tests/routers/test_health.py -v` — health router tests pass
- [ ] `TEST_DATABASE_URL=<pg_url> pytest tests/integration/ -m integration -v` — Postgres integration tests pass when a live DB is provided
- [ ] `GET /api/docs` returns Swagger UI with the full spec loaded


docs/api/openapi.yaml
```diff
@@ -0,0 +1,13654 @@
+openapi: 3.0.3
+info:
+  title: FastAPI
+  version: 1.0.0
+paths:
```
codewithwaleed, line 6: **[REQUIRED] Committed spec will drift from the live app immediately — needs CI enforcement**

The `openapi.yaml` is generated once and committed. The moment any router, schema, or response model changes (PRs #75–#78 all do this), the committed spec is stale. Consumers using this file as a contract will see misleading docs.

**Fix:** Add a CI step that regenerates the spec and fails if there are any changes:
```yaml
# .github/workflows/openapi-lint.yml
- name: Check OpenAPI spec is up to date
  run: |
    python scripts/export_openapi.py
    git diff --exit-code docs/api/openapi.yaml ||       (echo "OpenAPI spec is stale — run scripts/export_openapi.py and commit" && exit 1)
```

Without this, the spec becomes a liability rather than an asset within one sprint.

app/routers/api_docs.py
```diff
@@ -0,0 +1,34 @@
...
+from fastapi.responses import FileResponse, HTMLResponse
+
+router = APIRouter()
+
+# Repo root: app/routers/api_docs.py → parents[2]
+_OPENAPI_YAML = (
+    Path(__file__).resolve().parents[2] / "docs" / "api" / "openapi.yaml"
+)
+
+
```
codewithwaleed, line 19: **[REQUIRED] `/api/docs` and `/api/docs/openapi.yaml` are publicly accessible with no authentication**

```python
@router.get("/docs", include_in_schema=False)
async def api_docs() -> HTMLResponse:
```

Both endpoints have no `Depends(require_tenant)` or any auth guard. The full OpenAPI spec — including all endpoint paths, parameter names, schema shapes, and error codes — is exposed to unauthenticated requests. This gives attackers a complete map of your attack surface without needing a valid credential.

**Fix:** Either gate behind auth:
```python
@router.get("/docs", include_in_schema=False)
async def api_docs(user=Depends(require_tenant)) -> HTMLResponse:
    ...
```

Or verify the `RateLimitMiddleware` + `ApiKeyMiddleware` skip list does NOT include `/api/docs` (they currently do — the skip list bypasses rate limiting for docs, but auth is also bypassed because docs routes are unauthenticated by design). At minimum, add a note in the PR that this is an intentional public endpoint, and ensure the `_SKIP_EXACT` set in `rate_limit_middleware.py` is intentional.

tests/conftest.py
```diff
@@ -29,112 +31,278 @@
...
-TEST_DATABASE_URL = "sqlite://"
 
-engine = create_engine(
-    TEST_DATABASE_URL,
-    connect_args={"check_same_thread": False},
-    poolclass=StaticPool,  # Keep a single in-memory DB across threads
-)
+# ---------------------------------------------------------------------------
+# Custom markers
+# ---------------------------------------------------------------------------
```
codewithwaleed, line 53: **[REQUIRED] `_active_test_session` is a mutable module-level global — unsafe with parallel test workers**

```python
_active_test_session = None
```

`override_get_db` checks `if _active_test_session is not None: yield _active_test_session`. This global is mutated by the `db` fixture. With `pytest-xdist` (parallel workers) or if test collection order changes, two test modules can race on this global — one module\'s session leaks into another\'s requests, causing data cross-contamination and intermittent failures.

**Fix:** Use a threading local or fixture-scoped context variable instead:
```python
import contextvars
_active_test_session_var: contextvars.ContextVar = contextvars.ContextVar(
    "_active_test_session", default=None
)

def override_get_db():
    session = _active_test_session_var.get()
    if session is not None:
        yield session
        return
    db = TestingSessionLocal()
    try:
        yield db
    finally:
        db.close()
```

app/middleware/rate_limit_middleware.py
```diff
@@ -30,9 +30,8 @@
 
 _SKIP_EXACT = {
     "/",
```
codewithwaleed, line 34: **[WARNING] `/docs`, `/redoc`, `/openapi.json` removed from skip list but FastAPI built-in docs are now disabled**

```python
# Removed:
"/docs",
"/redoc",
"/openapi.json",
```

In `app/main.py` the built-in FastAPI docs are disabled (`docs_url=None, redoc_url=None, openapi_url=None`), so removing these from the skip list is correct — those routes no longer exist. However, if the built-in docs are ever re-enabled (e.g. in a dev environment override), requests to `/docs` will be rate-limited before the Swagger UI loads, causing confusing 429s.

Add a comment explaining why these were removed:
```python
# Built-in FastAPI docs disabled (docs_url=None in app factory) — these paths no longer exist.
# Custom docs served at /api/docs (already in _SKIP_EXACT above).
```

app/main.py
```diff
@@ -5,9 +5,8 @@
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.openapi.utils import get_openapi
-from fastapi.responses import HTMLResponse
-
 from app.api.api_v1.api import api_router
+from app.routers.api_docs import router as api_docs_router
 from app.routers.health import router as health_router
 from app.schemas.base import SuccessResponse
 from app.utils.response import create_success_response
```
codewithwaleed, line 12: **[WARNING] `docs_url=None, redoc_url=None, openapi_url=None` disables `/openapi.json` — breaks FastAPI automatic schema generation for tests**

```python
app = FastAPI(
    ...
    docs_url=None,
    redoc_url=None,
    openapi_url=None,
)
```

`tests/test_openapi_contract.py` likely calls `client.get("/openapi.json")` or uses `app.openapi()` to assert on schema shapes. Verify the contract tests do not rely on `/openapi.json` being served — if they do, they will silently return 404 instead of the schema, making assertions vacuously pass.

Also, disabling `/openapi.json` means FastAPI\'s `TestClient` schema introspection won\'t work in tests that use `response_model` assertions. Confirm all existing tests still pass after this change.

docs/api/openapi.yaml
```diff
@@ -0,0 +1,13654 @@
+openapi: 3.0.3
+info:
+  title: FastAPI
+  version: 1.0.0
```
codewithwaleed, line 5: **[SUGGESTION] OpenAPI `info.title` is `"FastAPI"` — should be the actual product name**

```yaml
info:
  title: FastAPI
  version: 1.0.0
```

The spec title shows `FastAPI` (the framework name) instead of the product name. Update `app/main.py`:
```python
app = FastAPI(
    title="TGS Voice Agent API",
    description="Multi-tenant SaaS Voice Agent Backend",
    version=settings.APP_VERSION,
    ...
)
```

Then re-run `scripts/export_openapi.py` to update the committed spec.

scripts/export_openapi.py
```diff
@@ -0,0 +1,43 @@
+"""Export the FastAPI OpenAPI spec to docs/api/openapi.yaml.
```
codewithwaleed, line 1: **[SUGGESTION] Export script should validate the generated spec before writing**

Add a schema validation step to catch malformed output early:
```python
# After generating, validate with openapi-spec-validator (pip install openapi-spec-validator)
from openapi_spec_validator import validate_spec
validate_spec(spec_dict)
```

This ensures the committed `openapi.yaml` is always a valid OpenAPI 3.x document and catches issues like missing `$ref` targets or invalid response schemas before they reach consumers.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #82: OpenAPI Spec, Workspace Security Docs & Integration Tests

**Overall: Good infrastructure sprint.** Committing the spec, exposing a dedicated Swagger endpoint, and hardening the SQLite cross-thread fixture are all solid improvements. Three issues require changes before merge — one critical (live spec drift), one security (unauthenticated spec exposure), one test reliability (shared mutable global). Two warnings and two suggestions also below.

app/routers/api_docs.py
Aliict: Gated /api/docs and /api/docs/openapi.yaml with HTTP Basic auth (API_DOCS_USERNAME / API_DOCS_PASSWORD); removed /api/docs from middleware skip lists; OpenAPI spec is no longer publicly accessible without credentials.

tests/conftest.py
Aliict: Replace the module-level _active_test_session global with a contextvars.ContextVar so each test worker/thread gets its own DB session and parallel runs don’t cross-contaminate data.

app/middleware/rate_limit_middleware.py
Aliict: Built-in /docs, /redoc, and /openapi.json are off (docs_url=None); custom API docs live at /api/docs, and rate limiting only runs on /api/v1/* paths anyway.

app/main.py
Aliict: Keep built-in /docs, /redoc, and /openapi.json off for production; openapi_url=None only removes those HTTP routes — schema generation still works via custom_openapi() / app.openapi() for tests and export.

docs/api/openapi.yaml
Aliict: Set OpenAPI info.title to Happy Assist.ai (with product description) in app/main.py and regenerate docs/api/openapi.yaml so the spec no longer shows the framework name.

scripts/export_openapi.py
Aliict: Validate exported OpenAPI with openapi-spec-validator before writing docs/api/openapi.yaml, normalizing FastAPI nullable schemas so malformed specs fail at export time.

PR #83, opened 2026-06-04
feat(vertex-ai): Vertex AI Gemini 2.5 Flash LLM integration for voice pipeline

Aliict:
## Summary

- **New `VertexGeminiService`** (`app/services/vertex_gemini_service.py`): async streaming Gemini 2.5 Flash via Vertex AI SDK using Application Default Credentials (ADC). Uses `generate_content_async` for true non-blocking token streaming — the previous `asyncio.to_thread(generate_content, stream=True)` approach buffered all tokens before returning, causing ~2.7s latency and single-chunk TTS delivery. Raises typed `VertexLlmError` (QUOTA / TIMEOUT / CONTENT_FILTER / UNKNOWN) with class-name + message-based fallback classification for test-runner compatibility.
- **`resolve_llm_runtime` + `llm_service_for_provider`** (`app/core/agent_runtime.py`): centralised LLM credential routing replacing ad-hoc provider selection in `generate_and_stream_response`. Gemini agents always use ADC — `model.api_key` is never decrypted for Gemini. Default temperature raised to 0.3 (`VOICE_LLM_DEFAULT_TEMPERATURE`).
- **`LlmPromptBuilder`** (`app/voice/llm_prompt_builder.py`): pure functions for Vertex content building — `prune_history_to_turns` (oldest pairs dropped first), `build_history_text`, `build_vertex_contents` (maps `(role, text)` history to Vertex `Content` objects with correct `user` / `model` roles).
- **`PipelineSession`** (`app/voice/pipeline_session.py`): per-call state dataclass holding shared references to `_tts_cancel` (barge-in event) and `_conversation_history_cache` so both access paths stay in sync.
- **Surgical `bidirectional_stream.py` changes**: new imports wired in `__init__`; `generate_and_stream_response` uses `resolve_llm_runtime` + `llm_service_for_provider`; Vertex path passes `cancel_event` + `conversation_history`; `VertexLlmError` caught separately with `VOICE_LLM_FALLBACK_MESSAGE` canned response — no cross-provider retry for Gemini agents.
- **Barge-in dead zone** (`VOICE_BARGE_IN_DEAD_ZONE_MS`, default 600 ms): Deepgram sends stale interims from the user's original speech at almost exactly the moment the first TTS audio frame is sent to Twilio — `_is_tts_playing` flips True and barge-in fires immediately, cutting every response before the user hears anything. The dead zone suppresses interim-triggered barge-in for 600 ms after `_tts_play_start_ts` is set, while final-transcript barge-in is unaffected.

## Files changed

| File | Type |
|------|------|
| `app/services/vertex_gemini_service.py` | New |
| `app/voice/llm_prompt_builder.py` | New |
| `app/voice/pipeline_session.py` | New |
| `tests/voice/test_vertex_llm.py` | New (30 tests) |
| `app/core/agent_runtime.py` | Modified |
| `app/core/config.py` | Modified |
| `app/routers/bidirectional_stream.py` | Modified (surgical) |
| `app/voice/tts_stream_mixin.py` | Modified |
| `requirements.txt` | Modified |
| `tests/voice/test_call_pipeline.py` | Modified |

## Pre-existing failures (not regressions)

`test_bidirectional_interim_regression.py::test_second_interim_does_not_start_second_llm` and `test_complete_final_regen_calls_generate_once` fail on the base branch before any changes — the `_empty_handler()` helper in that file does not set `_rag_prefetch_min_words` / `_rag_prefetch_min_confidence` which are accessed in `_maybe_process_interim`. Confirmed pre-existing via `git stash` before implementation.

## Test plan

- [ ] Run `pytest tests/voice/test_vertex_llm.py tests/core/test_agent_runtime.py tests/voice/test_call_pipeline.py` — 113 pass
- [ ] Verify `GOOGLE_APPLICATION_CREDENTIALS` or Workload Identity is set in staging/prod env
- [ ] Place a test call with a Gemini-configured agent and confirm audio response latency drops from ~2.7 s to <1.5 s (first audio frame)
- [ ] Trigger a barge-in after >600 ms of agent speaking — confirm it still cancels correctly
- [ ] Trigger a fast interim (<600 ms after TTS start) — confirm it is suppressed (`tts_dead_zone` log) and audio completes
- [ ] Confirm `gemini_service.embed_text` (RAG embeddings) still uses AI Studio — separate code path, unaffected

### DIFF
--- a/app/routers/bidirectional_stream.py
+++ b/app/routers/bidirectional_stream.py
@@ -119,9 +119,10 @@ from app.services.agent_service import agent_service
 from app.services.voice_logging_service import VoiceLoggingService
 from app.models.appointment import Appointment
 from app.services.transcript_service import transcript_service
-from app.services.gemini_service import gemini_service
 from app.services.openai_service import openai_service
 from app.services.groq_service import groq_service
+from app.services.vertex_gemini_service import VertexLlmError, vertex_gemini_service
+from app.core.agent_runtime import resolve_llm_runtime, llm_service_for_provider
 from app.services.rag_service import rag_service
 from app.services.credit_service import credit_service
 from app.services.twilio_service import twilio_service
@@ -181,6 +182,7 @@ from app.utils.eleven_tts_text import (
 from app.voice.booking_mixin import BookingMixin
 from app.voice.tts_stream_mixin import TtsStreamMixin
 from app.voice.call_control_mixin import CallControlMixin
+from app.voice.pipeline_session import PipelineSession
 
 router = APIRouter()
 
@@ -488,6 +490,14 @@ class BidirectionalStreamHandler(BookingMixin, TtsStreamMixin, CallControlMixin)
         # These are agent-level and don't change per-turn; caching them here
         # eliminates the 50-200ms parallel executor cost on every slow-path turn.
         asyncio.create_task(self._prefetch_kb_blocks_at_call_start())
+
+        # ── PipelineSession — per-call voice state container ─────────────────
+        # Holds references to the same objects as the existing handler attributes
+        # so both access paths stay in sync with no behaviour change.
+        self._pipeline_session = PipelineSession(
+            llm_cancel=self._tts_cancel,          # barge-in cancel event (shared ref)
+            history=self._conversation_history_cache,  # shared list ref
+        )
 
         # ── Voice Orchestration Layer ─────────────────────────────────────────
         # VoiceOrchestrator owns SttPipeline + TtsPipeline lifecycle and
@@ -860,8 +870,7 @@ class BidirectionalStreamHandler(BookingMixin, TtsStreamMixin, CallControlMixin)
                 else:
                     # Interim LLM already running with the correct transcript —
                     # let it finish without a second generation.
-                    pass
-                return
+                    return
             else:
                 self._turn_response_seed_text = ""
                 self._last_interim_text = ""
@@ -1402,6 +1411,9 @@ class BidirectionalStreamHandler(BookingMixin, TtsStreamMixin, CallControlMixin)
                 and self.call_session.call_metadata
                 and self.call_session.call_metadata.get("appointment_id")
             )
+            _llm_runtime = resolve_llm_runtime(self.agent)
+            _use_vertex_llm = (_llm_runtime.provider_slug or "").lower() == "gemini"
+            _vertex_kb_context = ""
             low_latency_fastpath = self._should_use_latency_fastpath(
                 user_text, booking_intent_turn
             )
@@ -1606,6 +1618,12 @@ class BidirectionalStreamHandler(BookingMixin, TtsStreamMixin, CallControlMixin)
                     )
                 except Exception:
                     history_text = ""
+
+            if _use_vertex_llm:
+                history_text = (
+                    "(Prior conversation turns are provided separately — "
+                    "maintain continuity; do not re-ask answered questions.)"
+                )
             
             booking_memory_block = self._build_booking_memory_block()
             follow_up_appt_block = self._build_follow_up_appointment_block()
@@ -1678,6 +1696,21 @@ class BidirectionalStreamHandler(BookingMixin, TtsStreamMixin, CallControlMixin)
                 "If the caller asks about the business, say that specific information is not "
                 "available to you right now and offer to help in another way."
             )
+
+            if _use_vertex_llm:
+                from app.voice.llm_prompt_builder import build_kb_context_for_vertex
+
+                _vertex_kb_context = build_kb_context_for_vertex(
+                    inbound_kb_docs_context_block,
+                    business_knowledge_block,
+                    empty_guard=_bk_block,
+                )
+                inbound_kb_docs_context_block = ""
+                _bk_block = (
+                    "# AUTHORITATIVE BUSINESS FACTS\n"
+                    "Business facts and inbound documents are attached as knowledge context "
+                    "with each user message — apply grounding rules using that context.\n"
+                )
 
             # Base prompt for phone conversations (voice-first, plain text only, no SSML)
             base_prompt = f"""# ROLE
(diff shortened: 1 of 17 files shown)


app/services/vertex_gemini_service.py
```diff
@@ -0,0 +1,249 @@
...
+from typing import Any, AsyncIterator
+
+from app.core.config import settings
+from app.core.logger import logger
+from app.voice.llm_prompt_builder import build_vertex_contents
+
+# Lazily imported so tests can patch before import resolution.
+_vertexai_initialized = False
+
+
```
codewithwaleed, line 24: **[REQUIRED] `_vertexai_initialized` is a module-level global — not safe for concurrent calls or test isolation**

```python
_vertexai_initialized = False

def _ensure_vertex_init() -> None:
    global _vertexai_initialized
    if _vertexai_initialized:
        return
    vertexai.init(...)
    _vertexai_initialized = True
```

Two problems:

1. **Race condition**: Two concurrent calls entering `_ensure_vertex_init()` simultaneously can both see `_vertexai_initialized = False` and call `vertexai.init()` twice. Vertex SDK may not be idempotent on double-init.

2. **Test pollution**: Once any test sets this to `True`, every subsequent test skips re-init. If a test patches `settings.GOOGLE_CLOUD_PROJECT_ID` to a different value, the already-initialized Vertex client uses the old project — silent wrong-project calls.

**Fix:** Use `threading.Lock` + double-checked locking:
```python
import threading
_vertex_init_lock = threading.Lock()
_vertexai_initialized = False

def _ensure_vertex_init() -> None:
    global _vertexai_initialized
    if _vertexai_initialized:
        return
    with _vertex_init_lock:
        if _vertexai_initialized:
            return
        vertexai.init(project=project, location=location)
        _vertexai_initialized = True
```

app/services/vertex_gemini_service.py
```diff
@@ -0,0 +1,249 @@
...
+        prompt: str,
+        system_prompt: str | None = None,
+        conversation_history: list[tuple[str, str]] | None = None,
+        kb_context: str | None = None,
+        model_name: str = "gemini-2.5-flash",
+        temperature: float = 0.3,
+        max_tokens: int = 100,
+        cancel_event: asyncio.Event | None = None,
+        # Ignored — Vertex uses ADC, not api_key. Accepted for interface compatibility.
+        api_key: str | None = None,
```
codewithwaleed, line 103: **[REQUIRED] `generate_content_async` is awaited and then iterated — but it returns an async generator, not a coroutine**

```python
async for response in await model.generate_content_async(
    contents,
    generation_config=generation_config,
    stream=True,
):
```

When `stream=True`, `generate_content_async()` returns an **async generator** directly — it is not a coroutine. Adding `await` here will raise `TypeError: object async_generator can't be used in 'await' expression` at runtime.

The correct pattern is:
```python
response_stream = model.generate_content_async(
    contents,
    generation_config=generation_config,
    stream=True,
)
async for response in response_stream:
    ...
```

This is a silent runtime failure — the `stream_text` call will crash on the first Gemini-agent call and fall back to the canned `VOICE_LLM_FALLBACK_MESSAGE`. Add a test that actually exercises the `await` path against the real SDK signature.

app/routers/bidirectional_stream.py
```diff
@@ -364,6 +366,14 @@ def __init__(
...
+        # starts playing.  Prevents Deepgram's stale interims (from the user's
+        # original speech still being processed) from immediately cutting the
+        # agent's response before the user has heard anything.
+        self._tts_play_start_ts: float = 0.0
+        self._barge_in_dead_zone_ms: float = float(
+            getattr(settings, "VOICE_BARGE_IN_DEAD_ZONE_MS", 600) or 600
+        )
         self._audio_samples_needed = max(
             4, int(getattr(settings, "VOICE_PICKUP_SAMPLE_WINDOW", 6) or 6)
         )
```
codewithwaleed, line 379: **[REQUIRED] Dead-zone check uses `time.perf_counter()` but `_tts_play_start_ts` is set in `tts_stream_mixin.py` — verify both use the same clock**

```python
# tts_stream_mixin.py:
self._tts_play_start_ts = _first_audio_ts   # _first_audio_ts = time.perf_counter()

# bidirectional_stream.py:
_in_dead_zone = (
    self._tts_play_start_ts > 0
    and (time.perf_counter() - self._tts_play_start_ts) * 1000 < self._barge_in_dead_zone_ms
)
```

Both sides use `time.perf_counter()` ✅ — this is consistent. However, `_tts_play_start_ts` is **never reset to 0.0** after the dead zone expires or after a barge-in. On the next utterance, if `_tts_play_start_ts` holds the timestamp from a previous turn, the dead-zone arithmetic produces a very large elapsed value, so `_in_dead_zone` is `False` — which happens to be correct. But it would be cleaner and safer to explicitly reset it:

```python
# In generate_and_stream_response, at the top of each new turn:
self._tts_play_start_ts = 0.0
```

This makes the intent explicit and prevents unexpected dead-zone suppression if a future code path changes when `_is_tts_playing` is set.

app/voice/llm_prompt_builder.py
```diff
@@ -0,0 +1,93 @@
...
+    parts = [p.strip() for p in (inbound_kb, business_kb) if p and p.strip()]
+    if parts:
+        return "\n\n".join(parts)
+    return (empty_guard or "").strip()
+
+
+def build_history_text(
+    history: list[tuple[str, str]],
+    max_turns: int = 20,
+) -> str:
```
codewithwaleed, line 63: **[WARNING] `build_vertex_contents` imports `vertexai` inside the function body — every call pays import resolution cost**

```python
def build_vertex_contents(...) -> list:
    try:
        from vertexai.generative_models import Content, Part
    except ImportError as exc:
        raise ImportError(...) from exc
```

For a function called on every voice turn (multiple times per call), this inner import is resolved each time. Python caches module imports so the overhead is small, but the `try/except ImportError` block runs on every call regardless.

**Fix:** Cache the import at module level with a lazy sentinel:
```python
_Content = None
_Part = None

def _load_vertex_models():
    global _Content, _Part
    if _Content is None:
        from vertexai.generative_models import Content, Part
        _Content, _Part = Content, Part
    return _Content, _Part
```

Or keep the in-function import but remove the `try/except` — if `google-cloud-aiplatform` is missing, the `ImportError` is clear enough without wrapping it.

app/services/vertex_gemini_service.py
```diff
@@ -0,0 +1,249 @@
...
+    if "DeadlineExceeded" in name or "timeout" in msg.lower() or "deadline" in msg.lower():
+        return VertexLlmError(f"Vertex timeout: {msg}", VertexLlmErrorType.TIMEOUT)
+    if "blocked" in msg.lower() or "safety" in msg.lower() or "filter" in msg.lower():
+        return VertexLlmError(f"Vertex content filter: {msg}", VertexLlmErrorType.CONTENT_FILTER)
+    return VertexLlmError(f"Vertex error: {msg}", VertexLlmErrorType.UNKNOWN)
+
+
+class VertexGeminiService:
+    """
+    Singleton service that streams Gemini 2.5 Flash responses via Vertex AI.
```
codewithwaleed, line 87: **[WARNING] `full_model_name = f"publishers/google/models/{model_name}"` is wrong for Vertex AI**

The correct Vertex AI model name format for Gemini is just `"gemini-2.5-flash"` or `"gemini-2.5-flash-preview-04-17"` — **not** prefixed with `publishers/google/models/`. That path format is used only in raw REST calls to the Model Garden API, not with the `google-cloud-aiplatform` Python SDK.

Using the wrong model name will cause a `404 Not Found` or `InvalidArgument` from Vertex AI on the first real call. The test suite mocks `GenerativeModel` so this bug is invisible in tests.

**Fix:**
```python
# Pass model_name directly — no publishers/ prefix needed with the SDK
model = GenerativeModel(
    model_name=model_name,  # e.g. "gemini-2.5-flash"
    **model_kwargs,
)
```

Verify against the [Vertex AI Python SDK docs](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini) — the SDK accepts the short model name.

codewithwaleed (changes requested):
## 🤖 AI Code Review — PR #83: Vertex AI Gemini 2.5 Flash + Barge-in Dead Zone

**Solid implementation.** The async streaming fix addresses a real 2.7s latency regression, the dead-zone guard solves a well-understood Deepgram timing race, and the `LlmPromptBuilder` pure-function refactor is a clean separation. Three required fixes and two warnings with specific line-level feedback below.

app/services/vertex_gemini_service.py
Aliict: Implemented thread-safe Vertex AI initialization using a lock-based double-checked pattern to prevent concurrent initialization issues.

app/services/vertex_gemini_service.py
Aliict: Remove incorrect await on generate_content_async(stream=True) — it returns an async generator, not a coroutine.

app/routers/bidirectional_stream.py
Aliict: Added reset of _tts_play_start_ts at the start of each turn to ensure correct and clean dead-zone handling.

app/voice/llm_prompt_builder.py
Aliict: Moved Vertex AI imports to a cached/module-level lazy loader to avoid repeated imports on each function call and improve efficiency.

app/services/vertex_gemini_service.py
Aliict: Use short Vertex SDK model names (gemini-2.5-flash) instead of publishers/google/models/ prefix

PR #88, opened 2026-06-09
feat(v2): bootstrap FastAPI v2 service with unified auth and API key dependency

Aliict:
## Summary

- Bootstraps the `/api/v2/` service layer running alongside `/api/v1/` with zero conflict
- Extends `ApiKeyMiddleware` to protect `/api/v2/*` routes using the same Redis-cached auth logic as v1
- Adds `get_current_workspace` to `app/api/deps.py` — async API-key dependency (reuses middleware state when already resolved, falls back to direct DB lookup)
- Adds `get_async_db` to `app/api/deps.py` — exposes the async SQLAlchemy session via `Depends()` for v2 route handlers
- Initialises the async DB connection pool in the lifespan startup hook; disposes cleanly on shutdown
- `app/api/v2/deps.py` is a thin re-export shim — no auth logic duplication; v2 routes import the same dependency functions as v1
- Adds `GET /api/v2/health` returning `{"status":"ok","version":"...","service":"fastapi-v2"}`
- Exposes Swagger UI at `/api/v2/docs` and a v2-filtered OpenAPI schema at `/api/v2/openapi.json`
- Pins all `requirements.txt` dependencies to exact installed versions
- Regenerates `docs/api/openapi.yaml` to include the new `/api/v2/health` route

## Changed files

| File | Change |
|---|---|
| `app/api/deps.py` | Added `get_async_db`, `get_current_workspace` (async, API-key-only) |
| `app/api/v2/deps.py` | Thin re-export shim — imports from `app/api/deps` only |
| `app/api/v2/routers/health.py` | `GET /api/v2/health` endpoint |
| `app/api/v2/api.py` | v2 router aggregation |
| `app/db/async_session.py` | Async engine + `async_sessionmaker`; `init_async_db` / `dispose_async_db` / `get_db` |
| `app/main.py` | Lifespan DB pool init/dispose; v2 router mount; `/api/v2/docs` + `/api/v2/openapi.json` |
| `app/middleware/api_key_middleware.py` | Extended to intercept `/api/v2/*`; added v2 public paths to skip list |
| `app/core/exception_handlers.py` | Minor handler update |
| `requirements.txt` | All packages pinned to exact versions |
| `docs/api/openapi.yaml` | Regenerated — includes `/api/v2/health` |
| `tests/api/v2/test_auth.py` | 6 tests: valid key, invalid key, missing key, workspace mismatch, inactive workspace, `require_tenant` via API key |
| `tests/api/v2/test_health.py` | Health endpoint smoke test |
| `tests/middleware/auth/test_api_key_middleware_unit.py` | Unit tests for updated middleware path guard |

## Auth architecture

```
Request to /api/v2/*
       ↓
ApiKeyMiddleware  (single middleware, shared v1 + v2)
  ├── public paths (health, docs, openapi.json) → pass through
  ├── x-api-key + x-workspace-id → resolve, Redis cache, set request.state
  ├── Authorization: Bearer      → validate JWT, load workspace, set request.state
  └── neither / invalid          → HTTP 401

       ↓  request.state populated

Route handler
  Depends(get_current_workspace)  — async API-key dep (app/api/deps.py)
  Depends(get_workspace)          — reads request.state.workspace
  Depends(require_tenant)         — API key or JWT user (unchanged from v1)
  Depends(require_admin / ...)    — role checks (unchanged from v1)
```

## Test plan

- [x] `pytest tests/api/v2/` — all passing
- [x] `pytest tests/middleware/auth/` — no regressions
- [x] `python scripts/export_openapi.py` — spec valid, `/api/v2/health` present
- [x] App starts via `uvicorn app.main:app --port 8000 --reload`


app/main.py
```diff
@@ -35,6 +37,7 @@
 async def lifespan(app: FastAPI):
     # ---- startup ----
     logger.info("Application startup initiated")
```
codewithwaleed, line 39: [REQUIRED] init_async_db() called before try/except — startup failure swallowed silently

init_async_db() is a sync call with no error handling. If DATABASE_URL is malformed or asyncpg is missing, it raises and FastAPI swallows the exception. The app starts in a broken state where every v2 route hits RuntimeError('Async DB not initialized').

Fix:

    try:
        init_async_db()
        logger.info('Async DB pool initialized')
    except Exception as exc:
        logger.critical('Failed to initialize async DB pool: %s', exc, exc_info=True)
        raise

Also consider making init_async_db() async so pool_pre_ping=True actually tests connectivity at startup rather than lazily on the first request.

app/api/deps.py
```diff
@@ -73,6 +79,68 @@ def get_workspace_api_key(request: Request) -> Workspace:
     return workspace
```
codewithwaleed, line 81: [REQUIRED] get_current_workspace short-circuit only covers API_KEY auth — JWT users pay a full DB round-trip on every v2 request

    if (
        existing is not None
        and get_auth_method(request) == AUTH_METHOD_API_KEY  # <-- JWT falls through
        and existing.id == workspace_id
    ):
        return existing

For JWT-authenticated requests the middleware already resolved the workspace and set request.state, but the check only short-circuits for AUTH_METHOD_API_KEY. JWT users always fall through to the full async DB lookup (API key hash + tenant join) even though the data is already in request.state.

Fix: remove the auth-method guard so the short-circuit fires for any already-resolved workspace:

    existing = get_workspace_from_request(request)
    if existing is not None and existing.id == workspace_id:
        if not existing.is_active:
            raise _UNAUTHORIZED_WORKSPACE
        return existing

If get_current_workspace is intentionally API-key-only, document that clearly and have it return 403 for JWT callers instead of silently doing a redundant DB lookup.

app/db/async_session.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import AsyncGenerator
+
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
+
+from app.core.config import settings
+from app.core.logger import logger
+from app.db.async_url import database_url_to_async
+
+_async_engine = None
+_AsyncSessionLocal: async_sessionmaker | None = None
```
codewithwaleed, line 13: [WARNING] Module-level _async_engine global is not thread-safe for parallel test workers or hot reload

    _async_engine = None
    _AsyncSessionLocal = None

Two problems:
1. pytest-xdist parallel workers race on init_async_db() with no lock.
2. During uvicorn --reload, a request in flight can call get_db() on a disposed _AsyncSessionLocal between dispose and re-init.

Fix: use threading.Lock with double-checked locking:

    import threading
    _init_lock = threading.Lock()

    def init_async_db() -> None:
        global _async_engine, _AsyncSessionLocal
        with _init_lock:
            if _async_engine is not None:
                return
            _async_engine = create_async_engine(
                database_url_to_async(settings.DATABASE_URL),
                pool_pre_ping=True,
            )
            _AsyncSessionLocal = async_sessionmaker(
                _async_engine, class_=AsyncSession, expire_on_commit=False
            )

requirements.txt
```diff
@@ -1,50 +1,51 @@
...
-uvicorn
-SQLAlchemy
-psycopg2-binary
-asyncpg
-alembic
-python-jose[cryptography]
-passlib[bcrypt]
-pydantic
-pydantic-settings
-pytest
```
codewithwaleed, line 11: [WARNING] redis==8.0.0b2 is a beta release pinned exactly — pip install will break if it is yanked from PyPI

    redis==8.0.0b2

Beta packages can be yanked; if that happens all CI and production installs fail with 'No matching distribution found'. Also, hard-pinning all transitive deps (greenlet==3.5.0 etc.) prevents security patches from auto-resolving.

Fix: use minimum-version constraints for application packages and drop the beta pin:

    redis>=4.6,<9.0       # 7.x stable is available
    greenlet>=3.0         # no need to pin transitive deps

Reserve exact pins only for packages with known compatibility breaks (bcrypt==4.1.2 is already a good example of a justified exact pin).

app/api/v2/routers/health.py
```diff
@@ -0,0 +1,14 @@
+from fastapi import APIRouter
+
+from app.core.config import settings
+
+router = APIRouter(tags=["v2"])
+
+
+@router.get("/health")
+async def health_check() -> dict:
```
codewithwaleed: [SUGGESTION] v2 health returns static dict — no DB liveness check means broken pools go undetected

    return {'status': 'ok', 'version': ..., 'service': 'fastapi-v2'}

A load balancer routing on this endpoint will keep sending traffic to a pod whose async DB pool is dead. The v2 service has an async DB pool — it should verify it.

Fix:

    from sqlalchemy import text
    from sqlalchemy.ext.asyncio import AsyncSession
    from app.db.async_session import get_db as get_async_db

    @router.get('/health')
    async def health_check(db: AsyncSession = Depends(get_async_db)) -> dict:
        try:
            await db.execute(text('SELECT 1'))
            db_status = 'ok'
        except Exception:
            db_status = 'error'
        return {
            'status': 'ok' if db_status == 'ok' else 'degraded',
            'version': settings.APP_VERSION,
            'service': 'fastapi-v2',
            'db': db_status,
        }

codewithwaleed (changes requested):
## AI Code Review — PR #88: v2 Bootstrap, Unified Auth, Async DB

Clean bootstrap. The middleware extension, thin re-export shim, async engine lifecycle, and middleware-state short-circuit are all well-structured. Four issues require changes before merge.

PR #91, opened 2026-06-10
Feat/db schema update

ahbilal220:
DB Schema V3 updates:

Column added: tenant.parent_workspace_id UUID NULLABLE FK -> tenant(id) ON DELETE SET NULL.
Column added: tenant.workspace_type TEXT CHECK IN ('agency','sub_account','standalone') DEFAULT 'standalone'.
New table: branding_configs (id UUID PK, workspace_id FK UNIQUE, logo_url TEXT,primary_colour CHAR(7), display_name TEXT, created_at, updated_at).
New table: pricing_configs (id UUID PK, workspace_id FK UNIQUE, per_minute_rate NUMERIC(10,4), markup_percent NUMERIC(5,2), created_at, updated_at).
New table: rbac_roles (id UUID PK, workspace_id FK, user_id FK, role TEXT CHECK IN ('admin','manager','config_only','read_only','billing_only'), created_at).
UNIQUE constraint: rbac_roles(workspace_id, user_id) — one role per user per workspace.
New table: usage_records (id UUID PK, workspace_id FK, call_id FK, billable_minutes NUMERIC(10,2), recorded_at TIMESTAMPTZ).
Index: workspaces(parent_workspace_id), rbac_roles(workspace_id, user_id), usage_records(workspace_id, recorded_at).
Seed script updated: create one agency workspace + one sub-account linked to it for local dev.


alembic/versions/537f4295714b_fixing_missing_tables.py
```diff
@@ -0,0 +1,865 @@
+"""Fixing missing tables
+
+Revision ID: 537f4295714b
+Revises: 20260429_user_products_link
+Create Date: 2026-05-04 15:33:01.180758
+
```
codewithwaleed: **Critical: this migration creates tables that already exist and creates a second migration head.**

`down_revision = "20260429_user_products_link"` branches off a much older point in history, while `166509c9a980` (the actual v3 migration in this PR) has `down_revision = "20260602_schema_v2"`. Running `alembic upgrade head` after this PR will fail with **multiple heads** since neither migration is an ancestor of the other.

Worse, this migration's `upgrade()` runs `op.create_table('plan', ...)`, `op.create_table('tenant', ...)`, `op.create_table('user', ...)`, `op.create_table('agent', ...)`, etc. - all of these tables already exist in the schema as of `20260429_user_products_link`/`20260602_schema_v2`. Applying this migration on any existing database will fail with `relation "tenant" already exists` (or similar) for every table.

This file looks like an accidental `alembic revision --autogenerate` against an empty/out-of-sync DB that got committed. Solution: delete this file entirely. If there is a real schema gap it should be a small targeted migration with idempotent `_has_table()` guards (see the pattern used in `20260602_schema_v2_completion.py`), chained after `20260602_schema_v2`.

app/models/usage_records.py
```diff
@@ -0,0 +1,24 @@
+from sqlalchemy import Column, DateTime, Numeric, Index, ForeignKey
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.orm import relationship
+from sqlalchemy.sql import func
+import uuid
+from app.db.base_class import Base
+
+
```
codewithwaleed, line 8: **FK references the wrong table name - `call_sessions` does not exist.**

`call_id = Column(UUID(...), ForeignKey("call_sessions.id", ondelete="SET NULL"), nullable=True)`

The actual table for `CallSession` is `callsession` (see `alembic/versions/20260425_add_slot_reservation_table.py` and `c4f1a2b3d5e6_add_calendar_tables.py`, both of which use `sa.ForeignKey("callsession.id")`).

The migration `166509c9a980_v3_sub_accounts_branding_rbac_roles.py` correctly uses `callsession.id` for the `usage_records.call_id` FK, but this ORM model uses `call_sessions.id` - the comment `# Adjusted based on your relation list` suggests this was guessed rather than verified. SQLAlchemy will raise `NoReferencedTableError` for `relationship("Tenant", back_populates="usage_records")` / on mapper configuration since the FK target table is wrong.

Solution: change to `ForeignKey("callsession.id", ondelete="SET NULL")` to match the migration and the existing `CallSession` model's table name.

alembic/versions/166509c9a980_v3_sub_accounts_branding_rbac_roles.py
```diff
@@ -0,0 +1,126 @@
...
+        ['workspace_id', 'user_id'], 
+        unique=True, 
+        postgresql_where=sa.text("deleted_at IS NULL")
+    )
+
+    # 5. Create usage_records table (Distinct from the old 'usagerecord')
+    op.create_table(
+        'usage_records',
+        sa.Column('id', sa.UUID(), nullable=False),
+        sa.Column('workspace_id', sa.UUID(), sa.ForeignKey('tenant.id', ondelete='CASCADE'), nullable=False),
```
codewithwaleed, line 94: **`uq_rbac_roles_workspace_user_active` is created via `op.create_index(..., unique=True, postgresql_where=...)` but is meant to enforce "one active role per user per workspace".**

This works functionally as a partial unique index, but the name `uq_*` (and the PR description's wording "UNIQUE constraint: rbac_roles(workspace_id, user_id)") implies a constraint, not an index - and there is also a separate non-unique `idx_rbac_roles_workspace_user` covering the exact same two columns (`op.create_index('idx_rbac_roles_workspace_user', 'rbac_roles', ['workspace_id', 'user_id'])`). The plain index is redundant: any query that would use `idx_rbac_roles_workspace_user` can use the partial unique index instead (Postgres can still use a partial index for `WHERE deleted_at IS NULL` lookups, and for general lookups the planner falls back fine since `(workspace_id, user_id)` is the leading column set).

Solution: drop `idx_rbac_roles_workspace_user` and keep only the partial unique index `uq_rbac_roles_workspace_user_active` - one less index to maintain and the same query coverage.

app/routers/agent.py
```diff
@@ -0,0 +1,443 @@
+from fastapi import APIRouter, Depends, HTTPException, status, Query
+from sqlalchemy.orm import Session
+from typing import Optional
+from app.schemas.agent import AgentCreate, AgentUpdate, AgentOut, AgentListResponse, LanguageEnum, VoiceTypeEnum
```
codewithwaleed, line 4: **This file (and `app/routers/resume.py`, `app/services/resume_rules_parser.py`, `app/services/tts_adapter.py`, `app/services/voice_twilio_utils.py`, `app/api/api_v1/endpoints/invite.py` - ~2000 lines combined) appears unrelated to the stated scope of this PR ("DB Schema V3 updates").**

None of these files are mentioned in the PR description, and they do not reference any of the new v3 tables/columns (`workspace_type`, `parent_workspace_id`, `branding_configs`, `pricing_configs`, `rbac_roles`, `usage_records`). This looks like a stale branch that picked up commits from another in-progress feature (resume parsing / agent CRUD / TTS adapter) during a rebase or merge.

Solution: rebase this branch onto the latest target branch and drop these unrelated files from the diff so this PR only contains the schema-v3 changes (migration, models, `seed_dev_workspace.py`, `docs/db/schema-v3.md`). Bundling unrelated large new routers into a schema-migration PR makes review and rollback much harder.

alembic/versions/166509c9a980_v3_sub_accounts_branding_rbac_roles.py
```diff
@@ -0,0 +1,126 @@
...
+def upgrade() -> None:
+    # 1. Modify the existing 'tenant' table (backward-compatible fields)
+    op.add_column('tenant', sa.Column('parent_workspace_id', sa.UUID(), sa.ForeignKey('tenant.id', ondelete='SET NULL'), nullable=True))
+    op.add_column('tenant', sa.Column('workspace_type', sa.String(), server_default='standalone', nullable=False))
+    
+    op.create_check_constraint(
+        'chk_tenant_workspace_type',
+        'tenant',
+        "workspace_type IN ('agency', 'sub_account', 'standalone')"
+    )
```
codewithwaleed, line 30: **`workspace_type` has no index despite being used for filtering (e.g. "list all sub-accounts of an agency", "find all standalone workspaces").**

Per project convention (CLAUDE.md: "Add indexes on foreign keys and any column used in `WHERE` filters on large tables"), if `workspace_type` will be used to filter the `tenant` table (which can grow large), it should have an index - possibly a partial index combined with `parent_workspace_id` for the agency -> sub-accounts lookup pattern, e.g. `CREATE INDEX ON tenant (parent_workspace_id) WHERE workspace_type = 'sub_account'`.

Not strictly blocking if `tenant` is expected to stay small (one row per customer workspace), but worth confirming and documenting in `docs/db/schema-v3.md`.

codewithwaleed (changes requested):
## AI Code Review - PR #91: Schema V3 (sub-accounts, branding, RBAC, usage)

The new v3 migration (`166509c9a980`) for tenant hierarchy/branding/pricing/rbac/usage tables is reasonable, but there are blocking issues with migration history and an FK mismatch, plus a large amount of unrelated code bundled into this PR that should not be here.

PR #92, opened 2026-06-10
feat(webhooks): Custom webhook system for call and batch events

Aliict:
## Summary

- **New models**: `WebhookEndpoint` + `WebhookDelivery` — stores tenant-configured outbound endpoints (secret encrypted at rest) and a full delivery audit log
- **New service**: `WebhookService` (CRUD, HMAC-SHA256 signing, HTTP delivery via `httpx`) + `fire_webhooks` background task function + `retry_webhook_delivery` ARQ job
- **New router** (`/api/v2/webhooks`): 5 endpoints — create, list, delete, test ping, delivery log; readonly role blocked on write operations; secret never returned in any response
- **Event triggers wired**:
  - `call.started` / `call.completed` / `call.failed` — fired via `BackgroundTasks` in `app/routers/voice.py`
  - `batch.completed` — fired via `asyncio.create_task` in `_maybe_complete_job` in `app/services/batch_call_worker_service.py`
- **Retry policy**: up to 4 total attempts (1 initial + 3 retries), delays of **1 min → 5 min → 30 min**, ARQ-backed using `_defer_until`
- **Migration**: `20260610_webhooks` — creates `webhookendpoint` and `webhookdelivery` tables, branches off `20260609_batch_calls`
- **Tests**: 23 tests in `tests/api/v2/test_webhooks.py` — endpoint CRUD, HTTPS URL validation, secret min-length, HMAC correctness, retry delay constants, delivery logging (200 / 500 / timeout)

## API Endpoints

| Method | Path | Description |
|--------|------|-------------|
| `POST` | `/api/v2/webhooks` | Register endpoint (HTTPS only, secret ≥ 16 chars) |
| `GET` | `/api/v2/webhooks` | List endpoints — secret never returned |
| `DELETE` | `/api/v2/webhooks/{id}` | Remove endpoint + cascade deliveries |
| `POST` | `/api/v2/webhooks/{id}/test` | Send test ping, returns delivery result |
| `GET` | `/api/v2/webhooks/{id}/deliveries` | Paginated delivery audit log |

## Event Payload Schema

```json
{
  "event": "call.completed",
  "workspace_id": "<uuid>",
  "timestamp": "<iso8601>",
  "data": { ... }
}
```

Supported events: `call.started`, `call.completed`, `call.failed`, `batch.completed`

## Security

- `X-Webhook-Signature: <hmac-sha256-hex>` header on every outgoing request
- Signing secret stored **encrypted** at rest (`encrypt_api_key` / `decrypt_api_key`) — reversible encryption is required for outbound HMAC signing
- Secret is never logged, never returned via API, never stored in delivery records

## Test Plan

- [x] `POST /api/v2/webhooks` — happy path returns 201 without secret field
- [x] HTTP URL rejected with 400
- [x] Secret under 16 chars rejected with 400
- [x] Missing auth returns 401
- [x] Readonly user blocked on POST/DELETE with 403
- [x] `GET /api/v2/webhooks` — list returns endpoints without secret
- [x] `DELETE /api/v2/webhooks/{id}` — 204 on success, 404 on missing
- [x] `POST /api/v2/webhooks/{id}/test` — ping delivery logged and returned
- [x] `GET /api/v2/webhooks/{id}/deliveries` — paginated log returned
- [x] HMAC signature matches `hmac.new(secret, payload, sha256).hexdigest()`
- [x] Retry delays are `[1, 5, 30]` minutes
- [x] ARQ `_defer_until` passed correctly on retry schedule
- [x] 500 response logged as `status=failed` with `http_status=500`
- [x] Timeout logged as `status=failed` with `response_body="timeout"`
- [x] 200 response logged as `status=delivered`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/app/models/webhook.py
+++ b/app/models/webhook.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import uuid
+
+from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Index, Integer, Text
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+from sqlalchemy.orm import relationship
+from sqlalchemy.sql import func
+
+from app.db.base_class import Base
+
+
+class WebhookEndpoint(Base):
+    """Tenant-configured outbound webhook endpoint."""
+
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    workspace_id = Column(
+        UUID(as_uuid=True),
+        ForeignKey("tenant.id", ondelete="CASCADE"),
+        nullable=False,
+    )
+    url = Column(Text, nullable=False)
+    # Signing secret stored encrypted at rest (encrypt_api_key / decrypt_api_key).
+    # Not a one-way hash — must be reversible for outbound HMAC signing.
+    secret = Column(Text, nullable=False)
+    is_active = Column(Boolean, nullable=False, default=True, server_default="true")
+    created_at = Column(
+        DateTime(timezone=True), nullable=False, server_default=func.now()
+    )
+
+    workspace = relationship("Tenant")
+    deliveries = relationship(
+        "WebhookDelivery", back_populates="endpoint", cascade="all, delete-orphan"
+    )
+
+    __table_args__ = (
+        Index("ix_webhookendpoint_ws", "workspace_id"),
+    )
+
+
+class WebhookDelivery(Base):
+    """Log of every outbound webhook delivery attempt."""
+
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    endpoint_id = Column(
+        UUID(as_uuid=True),
+        ForeignKey("webhookendpoint.id", ondelete="CASCADE"),
+        nullable=False,
+    )
+    event_type = Column(Text, nullable=False)
+    payload = Column(JSONB, nullable=False)
+    # "delivered" | "failed" | "retrying"
+    status = Column(Text, nullable=False)
+    http_status = Column(Integer, nullable=True)
+    response_body = Column(Text, nullable=True)
+    attempt_count = Column(Integer, nullable=False, default=1, server_default="1")
+    last_attempted_at = Column(
+        DateTime(timezone=True), nullable=False, server_default=func.now()
+    )
+    created_at = Column(
+        DateTime(timezone=True), nullable=False, server_default=func.now()
+    )
+
+    endpoint = relationship("WebhookEndpoint", back_populates="deliveries")
+
+    __table_args__ = (
+        Index("ix_webhookdelivery_ep", "endpoint_id"),
+    )
--- a/app/routers/voice.py
+++ b/app/routers/voice.py
@@ -1,4 +1,4 @@
-from fastapi import APIRouter, Request, HTTPException, Query, Depends, status, UploadFile, File, Form
+from fastapi import APIRouter, BackgroundTasks, Request, HTTPException, Query, Depends, status, UploadFile, File, Form
 from fastapi.responses import HTMLResponse, StreamingResponse, JSONResponse
 from sqlalchemy.orm import Session, joinedload
 from typing import Optional
@@ -247,6 +247,7 @@ _TWILIO_TO_INTERNAL_STATUS: dict[str, str] = {
 @router.post("/webhook/call-events", response_class=HTMLResponse, include_in_schema=False)
 async def handle_call_events_webhook(
     request: Request,
+    background_tasks: BackgroundTasks,
     agentId: Optional[str] = Query(None),
     userId: Optional[str] = Query(None),
     callSessionId: Optional[str] = Query(None),
@@ -546,7 +547,7 @@ async def handle_call_events_webhook(
         if call_status == "initiated" and direction == "outbound-api":
             # Call has been initiated - just log and return empty response
             logger.info(f"Call initiated - SID: {call_sid}")
-            
+
             # Broadcast call initiated event (non-blocking - fire and forget)
             if call_session:
                 try:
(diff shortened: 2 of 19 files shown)


app/schemas/webhook.py
```diff
@@ -0,0 +1,50 @@
...
+
+class WebhookEndpointCreate(BaseModel):
+    url: AnyHttpUrl
+    secret: str = Field(min_length=16)
+
+    @field_validator("url")
+    @classmethod
+    def must_be_https(cls, v: AnyHttpUrl) -> AnyHttpUrl:
+        if str(v).startswith("http://"):
+            raise ValueError("Webhook URL must use HTTPS")
```
codewithwaleed, line 21: **SSRF: URL validation only checks the scheme, not the target host.**

`must_be_https` rejects `http://` but a tenant can still register `https://169.254.169.254/latest/meta-data/`, `https://localhost:6379`, `https://10.0.0.5/internal-admin`, etc. Both `send_test_ping` and `fire_webhooks` will then have the FastAPI server make an outbound `httpx` POST to that address with the tenant's signed payload - a classic SSRF vector that can reach cloud metadata endpoints, internal services, or other tenants' infrastructure on the same network.

**Proposed fix** - resolve the hostname and reject private/loopback/link-local/metadata ranges, both at creation time and again at delivery time (DNS can change between registration and delivery - TOCTOU):

```python
import ipaddress
import socket
from urllib.parse import urlparse

_BLOCKED_HOSTS = {"localhost", "169.254.169.254"}

def assert_public_url(url: str) -> None:
    host = urlparse(url).hostname
    if not host or host.lower() in _BLOCKED_HOSTS:
        raise ValueError("Webhook URL host is not allowed")
    try:
        addrs = {info[4][0] for info in socket.getaddrinfo(host, None)}
    except socket.gaierror as exc:
        raise ValueError("Webhook URL host could not be resolved") from exc
    for addr in addrs:
        ip = ipaddress.ip_address(addr)
        if (
            ip.is_private
            or ip.is_loopback
            or ip.is_link_local
            or ip.is_reserved
            or ip.is_multicast
        ):
            raise ValueError("Webhook URL resolves to a non-public address")
```

Call `assert_public_url(str(v))` from `must_be_https` (creation time) and again from `WebhookService._attempt_delivery` right before the `httpx` call (delivery time), raising into the existing `except Exception` so a blocked delivery is logged as `status="failed"` rather than crashing the background task.

app/models/webhook.py
```diff
@@ -0,0 +1,68 @@
...
+
+class WebhookEndpoint(Base):
+    """Tenant-configured outbound webhook endpoint."""
+
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    workspace_id = Column(
+        UUID(as_uuid=True),
+        ForeignKey("tenant.id", ondelete="CASCADE"),
+        nullable=False,
+    )
```
codewithwaleed, line 21: **New `secret` column uses the legacy JWT-based `encrypt_api_key` / `decrypt_api_key` (from `app.core.security`, backed by `jose.jwt`), not pgcrypto.**

PR #81 / `20260602_schema_v2_completion.py` migrated `agent.encrypted_elevenlabs_api_key` away from this exact JWT-based scheme to `pgp_sym_encrypt`/`pgp_sym_decrypt` specifically because the JWT format was being deprecated for at-rest secrets. This PR introduces a brand-new column using the scheme that's actively being migrated away from elsewhere in the codebase - it will need its own re-encryption migration shortly after merging.

**Solution**: store `secret` using `pgp_sym_encrypt(:pt, :key)` / `pgp_sym_decrypt` with `ELEVENLABS_ENCRYPTION_KEY` (or a new dedicated `WEBHOOK_SECRET_ENCRYPTION_KEY` setting) the same way `_remigrate_elevenlabs_keys` does, via `db_encryption.py` helpers if they exist, instead of `encrypt_api_key`/`decrypt_api_key`.

app/services/webhook_service.py
```diff
@@ -0,0 +1,379 @@
...
+                endpoint_id=endpoint.id,
+                event_type=event_type,
+                payload=payload,
+                status="delivered" if delivered else "failed",
+                http_status=http_status,
+                response_body=response_body,
+                attempt_count=1,
+                last_attempted_at=now,
+            )
+            self._db.add(delivery)
```
codewithwaleed, line 231: **`fire_webhooks` does not filter by `WebhookEndpoint.is_active` consistently with retries, and a slow/unreachable endpoint blocks delivery to the next endpoint.**

Each `await svc._attempt_delivery(...)` in the `for endpoint in endpoints` loop runs sequentially with a 5s timeout. With multiple endpoints registered for a workspace, one slow endpoint adds up to 5s of latency per endpoint to this background task before the next one is even attempted. Since this already runs as a `BackgroundTasks`/`asyncio.create_task` job (not on the request path), this is not blocking, but for workspaces with many endpoints consider `asyncio.gather` to deliver to all endpoints concurrently:

```python
await asyncio.gather(
    *(_deliver_to_endpoint(db, ep, payload, event_type) for ep in endpoints),
    return_exceptions=True,
)
```

Not blocking for this PR, just worth tracking if the number of endpoints per workspace grows.

app/routers/voice.py
```diff
@@ -563,7 +564,25 @@ async def handle_call_events_webhook(
...
+                            "from_number": from_number,
+                            "to_number": to_number,
+                        },
+                    )
+                except Exception as _wh_exc:
+                    logger.warning("call.started webhook fire failed: %s", _wh_exc)
+
             return HTMLResponse("", media_type="application/xml")
         
         elif call_status == "ringing" and direction == "outbound-api":
```
codewithwaleed, line 588: **`call.completed` and `call.failed` webhook payloads include raw `from_number`/`to_number` (PII - phone numbers) sent to a tenant-supplied, now-unvalidated URL.**

Combined with the SSRF gap above, an attacker who registers a malicious endpoint and lets a real call complete would receive caller/callee phone numbers plus call duration over HTTPS to a server they control - which is by design for a webhook integration (tenant's own data), so this is fine *once* the URL-validation fix in `app/schemas/webhook.py` lands. Just flagging the dependency: don't merge the webhook payload changes ahead of the SSRF fix, since at that point any tenant could pivot this into reading back internal-network responses (not just their own call data) by pointing the endpoint at an internal service and triggering a call event.

codewithwaleed (changes requested):
## AI Code Review - PR #92: Custom Webhook System

Solid feature overall - good test coverage, secret never returned, retry/backoff via ARQ is well structured. Two issues should be addressed before merge: an SSRF gap on user-supplied webhook URLs, and using legacy JWT-based encryption for the new secret column instead of the pgcrypto path this codebase is migrating to.

PR #93, opened 2026-06-11
feat(kb): pgvector knowledge base ingestion pipeline

Aliict:
## Summary

- **Replace Pinecone with pgvector** — RAG embeddings now stored in PostgreSQL using `vector(1536)` with a cosine HNSW index (`m=16, ef_construction=64`). No external vector DB dependency.
- **New schema** — three tables added via Alembic migration: `knowledgebase`, `kbfile`, `kbchunk`. Previous `knowledgebasedocument` / `knowledgebasechunk` tables dropped and recreated.
- **File ingestion endpoint** (`POST /{kb_id}/file`) — accepts PDF, DOCX, TXT ≤ 50 MB. Streams to GCS, creates a `KbFile` record, enqueues `kb_ingestion_task` on the existing ARQ worker. Returns 202 `{job_id, file_id}`.
- **Text ingestion endpoint** (`POST /{kb_id}/text`) — synchronous; tiktoken cl100k_base chunking (800 tok / 100 overlap / 50 min), OpenAI `text-embedding-ada-002` embeddings, inserts `KbChunk` rows. Returns 201.
- **Status polling** (`GET /{kb_id}/files/{file_id}/status`) — returns `{status, chunk_count, error_message}`.
- **Auth** — all KB endpoints switched from `require_admin` to `require_tenant` (valid for both JWT users and M2M API keys).
- **Model cleanup** — removed explicit `__tablename__` from 7 models (`KnowledgeBase`, `KbChunk`, `KbFile`, `BrandingConfig`, `PricingConfig`, `RbacRole`, `UsageRecord`); names now auto-generated by the declarative base (`cls.__name__.lower()`). Single Alembic migration renames all affected tables and merges the previously diverged migration heads into one.
- **`usage_records.py` consolidated** — plural file deleted; `usage_record.py` now holds the per-call billing audit model (`UsageRecord`). `Subscription.usage_records` back-reference removed (FK no longer exists on the new model).
- **Body limit raised** — global `BodyLimitMiddleware` cap bumped from 10 MB to 52 MB to allow the 50 MB file upload path to reach the per-endpoint validation.

## New files

| File | Purpose |
|---|---|
| `app/models/kb_file.py` | `KbFile` ORM model |
| `app/services/kb_ingestion_service.py` | PDF/DOCX/TXT extraction, tiktoken chunking, OpenAI embedding, GCS upload |
| `alembic/versions/20260611_replace_kb_tables_pgvector.py` | Drop old KB tables, enable pgvector, create new schema + HNSW index |
| `alembic/versions/20260611_rename_tables_to_base_class_names.py` | Rename 7 tables to base-class names, merge heads |
| `tests/routers/test_knowledge_base.py` | 12 tests covering all new endpoints + chunker unit tests |

## Test plan

- [x] `pytest tests/routers/test_knowledge_base.py` — 12 tests, all green
- [x] `pytest tests/routers/test_health.py tests/routers/test_scheduled_calls.py` — 10 existing tests, all green
- [x] `alembic upgrade head` applies cleanly (single head: `20260611_rename_tables_base_class`)
- [x] Upload a real PDF via `POST /{kb_id}/file`, poll status until `ready`, verify `chunk_count > 0`
- [x] Ingest text via `POST /{kb_id}/text`, verify chunks returned in RAG retrieve preview
- [x] M2M API key can access all KB endpoints (require_tenant covers both JWT and API key)


app/routers/knowledge_base.py
```diff
@@ -1,189 +1,399 @@
...
+    job_id = str(uuid.uuid4())
+    pool = get_arq_pool()
+    if pool:
+        try:
+            asyncio.get_event_loop().run_until_complete(
+                pool.enqueue_job("kb_ingestion_task", str(file_id))  # type: ignore[union-attr]
+            )
+        except Exception as e:
+            logger.warning("ARQ enqueue failed for file_id=%s: %s", file_id, e)
+    else:
```
codewithwaleed, line 196: **`asyncio.get_event_loop().run_until_complete(...)` inside a sync `def` route handler will break or behave unpredictably.**

FastAPI runs sync `def` endpoints in a worker thread via `anyio`/threadpool. In Python 3.10+, `asyncio.get_event_loop()` on a non-main thread with no running loop raises `DeprecationWarning` and, depending on version, `RuntimeError: There is no current event loop`. Even when it doesn't raise, calling `run_until_complete` on a freshly created loop per-request is wasteful and can leak loops/connections (`get_arq_pool()`'s underlying redis connection is bound to whatever loop created it).

**Solution** - make the endpoint `async def` (it does no CPU-heavy work itself - GCS upload and the synchronous body read are the only blocking parts) and `await` the enqueue directly:

```python
@router.post("/{kb_id}/file", ...)
async def upload_kb_file(
    kb_id: uuid.UUID,
    file: UploadFile = File(...),
    user=Depends(require_tenant),
    db: Session = Depends(get_db),
):
    ...
    pool = get_arq_pool()
    if pool:
        try:
            await pool.enqueue_job("kb_ingestion_task", str(file_id))
        except Exception as e:
            logger.warning("ARQ enqueue failed for file_id=%s: %s", file_id, e)
```

This matches `ingest_kb_text`, which is already correctly `async def` and awaits the ingestion pipeline directly.

app/models/knowledge_base_chunk.py
```diff
@@ -1,33 +1,46 @@
...
-    - stable chunk references for auditing/debugging
+    The embedding column is stored as TEXT (JSON array string) in Python/SQLite,
+    and as vector(1536) in PostgreSQL — cast happens at query time via raw SQL.
     """
 
     id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, index=True)
-    document_id = Column(UUID(as_uuid=True), ForeignKey("knowledgebasedocument.id"), nullable=False, index=True)
-
-    chunk_index = Column(Integer, nullable=False, index=True)
-    vector_id = Column(String(300), nullable=False, unique=True, index=True)
```
codewithwaleed, line 25: **`embedding` is declared as `Column(Text, ...)` but the migration creates the actual Postgres column as `vector(1536)`.**

`pgvector==0.3.6` was added to `requirements.txt` in this PR but is never imported - the model still uses a plain `Text` column with a comment explaining the type mismatch is intentional ('actual column type is vector(1536) in Postgres'). Relying on Postgres to coerce a TEXT-bound parameter (`json.dumps(embedding)` -> `'[0.1, 0.2, ...]'`) into `vector` on every INSERT works only because of pgvector's implicit text->vector input cast, and breaks `db.refresh()`/ORM round-trips that expect the Python value to stay a `str` (a freshly-read row will come back as pgvector's own text representation, not necessarily byte-identical JSON).

**Solution** - use the `pgvector` package's SQLAlchemy integration, which is already a dependency:

```python
from pgvector.sqlalchemy import Vector

embedding = Column(Vector(1536), nullable=True)
```

and pass `embedding=embedding_list` (a `list[float]`, not `json.dumps(...)`) from `kb_ingestion_service.py`. This gives correct type round-tripping and lets `c.embedding <=> CAST(:vec AS vector)` in `rag_service.py` drop the redundant `::vector` cast on the column side.

app/main.py
```diff
@@ -210,8 +210,8 @@ def custom_openapi():
 # 3. PII-safe request logging
 app.add_middleware(PiiLoggingMiddleware)
 
-# 4. Body size limit (10 MB)
-app.add_middleware(BodyLimitMiddleware)
+# 4. Body size limit — 52 MB to allow KB file uploads (max 50 MB per endpoint)
```
codewithwaleed: **Global request body limit raised from 10 MB to 52 MB for every endpoint, not just the new KB file-upload route.**

`BodyLimitMiddleware` wraps the whole app, so every existing endpoint (webhooks, batch CSV upload, auth, etc.) now accepts request bodies up to 52 MB. This widens the memory-exhaustion surface for endpoints that were previously protected by the 10 MB cap and have no reason to accept large bodies (e.g. JSON API endpoints).

**Solution** - keep the global default at 10 MB and apply the larger limit only to the KB upload route, e.g. via a per-route dependency or by reading `request.url.path` in the middleware:

```python
app.add_middleware(BodyLimitMiddleware, max_bytes=10 * 1024 * 1024)
```
and in `body_limit_middleware.py`:
```python
_LARGE_BODY_PREFIXES = ("/api/v1/knowledge-base",)  # adjust to actual prefix
...
limit = 52 * 1024 * 1024 if request.url.path.startswith(_LARGE_BODY_PREFIXES) else self.max_bytes
```

alembic/versions/20260611_rename_tables_to_base_class_names.py
```diff
@@ -0,0 +1,53 @@
+"""Rename tables to match SQLAlchemy base-class auto-naming (cls.__name__.lower()).
+
+Also merges the two diverged heads:
+  - 166509c9a980  (branding_configs / pricing_configs / rbac_roles / usage_records)
+  - 20260611_kb_pgvector  (knowledge_bases / kb_files / kb_chunks)
+
+Renames:
+  branding_configs  → brandingconfig
+  pricing_configs   → pricingconfig
```
codewithwaleed, line 9: **Merge migration's `down_revision` tuple `("166509c9a980", "20260611_kb_pgvector")` does not account for `537f4295714b_fixing_missing_tables.py` introduced in PR #91.**

PR #91 (still open / under review) adds `537f4295714b` as a *third* divergent head with `down_revision = "20260429_user_products_link"`, which also recreates tables (`tenant`, `user`, `agent`, etc.) that already exist. If both PR #91 and this PR merge as-is, `alembic upgrade head` will have 3 unmerged heads and this merge migration will not resolve them, plus `537f4295714b` will fail on its own (duplicate table creation).

**Solution** - this isn't something to fix in PR #93 itself, but merge order matters: either land the fix for PR #91 (delete `537f4295714b_fixing_missing_tables.py`, per that PR's review) before this one, or extend this merge migration's `down_revision` tuple to include all three heads once PR #91's migration situation is resolved. Flagging here so it isn't missed when both branches are rebased onto the same target.

codewithwaleed (changes requested):
## AI Code Review - PR #93: pgvector Knowledge Base Pipeline

Big, well-organized swap from Pinecone to pgvector with good test coverage. A few issues before merge: a sync-route asyncio misuse on the upload path, an ORM/DB type mismatch on the embedding column, a global body-limit increase that widens the DoS surface for unrelated endpoints, and a migration-head dependency gap with PR #91.

PR #94, opened 2026-06-11
feat(kb): real-time KB retrieval layer — per-call-turn context injection (BE1-S7)

Aliict:
## Summary

Implements the retrieval side of the knowledge base pipeline (BE1-S7). When a call flow has `knowledge_base_ids` attached, relevant KB chunks are retrieved, ranked, and injected into the LLM system prompt before every conversation turn — without blocking the voice latency budget.

### What changed

- **`app/services/kb_retrieval_service.py`** — Core retrieval service:
  - Embeds caller transcript via `text-embedding-ada-002`, cached in Redis for 300 s (sha256 key)
  - Full result cache: `sha256(transcript + ':' + ':'.join(sorted(kb_ids)))`, TTL 300 s
  - Parallel per-KB pgvector cosine queries via `asyncio.gather(..., return_exceptions=True)`
  - Merges results across KBs, sorts by score descending, keeps global top 5
  - Fails open on all errors (embedding failure, DB failure, timeout) — call is never blocked

- **`app/voice/conversation_orchestrator.py`** — KB retrieval wired into call turn loop:
  - Checks `flow.knowledge_base_ids` before every LLM call
  - Wrapped in `asyncio.wait_for(..., timeout=0.45)` to stay within 450 ms voice budget
  - Injects context block after system prompt, before conversation history:
    ```
    --- KNOWLEDGE BASE CONTEXT ---
    {chunk_1}
    ---
    {chunk_2}
    --- END CONTEXT ---
    ```
  - Structured latency logging per turn: `kb_retrieval latency_ms=… kb_count=… call_sid=…`

- **`app/routers/bidirectional_stream.py`** — Loads `CallFlow` record into handler on session init so `conversation_orchestrator` can read `knowledge_base_ids` without an extra DB round-trip per turn

- **`app/routers/knowledge_base.py`** — New `GET /{kb_id}/search?q={query}&limit={n}` endpoint for direct KB search; returns `[{content, score, metadata}]`

- **`app/schemas/knowledge_base.py`** — Added `KbSearchResultItem` and `KbSearchResponse` schemas

- **`app/models/call_flow.py`** — Added `knowledge_base_ids` JSONB column (`default=list`)

- **`app/utils/redis_client.py`** — Shared lazy-init async Redis client (returns `None` gracefully when `REDIS_URL` is unset)

- **`alembic/versions/20260611_add_kb_ids_to_callflow.py`** — Migration: adds `knowledge_base_ids JSONB` to `callflow` table (applied ✅)

- **`alembic/versions/20260611_cleanup_orphan_tables.py`** — Companion migration: stamps orphan tables that existed in DB ahead of Alembic tracking

- **`.github/workflows/openapi-lint.yml`** — CI check: regenerates `docs/api/openapi.yaml` on every PR touching `app/**` and fails if the committed spec is stale

### Tests

12 new tests in `tests/services/test_kb_retrieval_service.py`, all passing:

| Test | Covers |
|---|---|
| `test_retrieve_returns_block_when_kb_attached` | KB attached to flow → context block in output |
| `test_retrieve_redis_cache_hit_skips_embedding` | Cache hit → embedding function not called |
| `test_retrieve_partial_failure_continues` | One KB errors → other KB results still returned |
| `test_retrieve_no_kb_ids_returns_empty` | No KB IDs → `("", 0.0)` fast path |
| `test_retrieve_embedding_failure_returns_empty` | Embedding failure → empty string, call not blocked |
| `test_retrieve_top5_merge_across_kbs` | Multi-KB merge, global top 5 kept |
| `test_format_kb_context_block_spec_format` | Correct header/footer/separator format |
| `test_format_kb_context_block_empty` | Empty input → empty string |
| `test_format_kb_context_block_single_chunk` | Single-chunk edge case |
| `test_search_endpoint_returns_results` | `GET /kb/{id}/search` happy path |
| `test_search_endpoint_404_unknown_kb` | Unknown KB ID → 404 |
| `test_search_endpoint_400_no_openai_key` | Missing `OPENAI_API_KEY` → 400 |

### Test plan

- [x] Run `pytest tests/services/test_kb_retrieval_service.py -v` — all 12 tests green
- [x] Run full suite `pytest tests/ -v` — no regressions
- [x] Attach a call flow with `knowledge_base_ids` set and make a voice call; verify KB context block appears in LLM prompt logs
- [x] Hit `GET /api/v1/kb/{kb_id}/search?q=<query>` with a real KB — verify ranked results returned
- [x] Confirm `alembic upgrade head` runs cleanly on a fresh DB (`knowledge_base_ids` column present in `callflow`)


alembic/versions/20260611_cleanup_orphan_tables.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def upgrade() -> None:
+    for tbl in _ORPHAN_TABLES:
+        op.execute(f"DROP TABLE IF EXISTS {tbl} CASCADE")
+
+    # The old subscription-quota `usagerecord` (subscription_id, month, year,
+    # calls_used, agents_created) conflicts with the rename target. Drop it first,
+    # then rename the billing-audit `usage_records` → `usagerecord`.
+    op.execute("DROP TABLE IF EXISTS usagerecord CASCADE")
```
codewithwaleed, line 46: **Critical: this drops the billing-audit `usagerecord` table that the previous migration just created, resulting in permanent data loss on a fresh DB / CI run.**

This migration's `down_revision` chain runs *after* `20260611_rename_tables_to_base_class_names.py`, which (per this same PR's diff) now does:
```python
op.execute("DROP TABLE IF EXISTS usagerecord CASCADE")
op.rename_table("usage_records", "usagerecord")
```
So by the time this migration runs on a DB that applies migrations in order, `usage_records` no longer exists (it was renamed) and `usagerecord` already holds the billing-audit data.

Then this migration does:
```python
op.execute("DROP TABLE IF EXISTS usagerecord CASCADE")  # <-- drops the billing-audit table just renamed!
op.execute("ALTER TABLE IF EXISTS usage_records RENAME TO usagerecord")  # no-op, usage_records doesn't exist
```
Net effect on a clean apply: `usagerecord` (billing-audit, with real data) is dropped and never recreated.

The docstring's premise ("the rename migration was stamped without executing") only holds for *already-deployed* environments where that migration was stamped, not for fresh installs/CI/new environments where both migrations actually execute — those will lose the table.

**Solution:** Make this idempotent regardless of whether the rename migration's body actually ran. E.g. only drop the legacy quota-shaped `usagerecord` if it still has the *old* schema:
```python
op.execute("""
    DO $$
    BEGIN
        IF EXISTS (SELECT 1 FROM information_schema.columns
                   WHERE table_name = 'usagerecord' AND column_name = 'subscription_id')
           AND NOT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'usage_records')
        THEN
            -- already renamed by the prior migration on this DB; nothing to do
            NULL;
        ELSIF EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'usage_records') THEN
            DROP TABLE IF EXISTS usagerecord CASCADE;
            ALTER TABLE usage_records RENAME TO usagerecord;
        END IF;
    END $$;
""")
```
Or simpler: remove the duplicate drop/rename from *one* of the two migrations entirely — pick whichever migration is guaranteed to run in every environment (probably this cleanup one, since the rename migration may already be stamped in prod) and drop the corresponding block from the other.

alembic/versions/20260611_rename_tables_to_base_class_names.py
```diff
@@ -28,7 +29,13 @@ def upgrade() -> None:
```
codewithwaleed, line 27: Same `usagerecord` drop/rename as in the new `20260611_cleanup_orphan_tables.py` migration (down_revision-chained after this one). Running both in sequence on a fresh DB drops the renamed billing-audit table a second time and loses its data — see the comment on `20260611_cleanup_orphan_tables.py` for details/solution. Pick one migration to own this rename and remove it from the other.

app/services/kb_retrieval_service.py
```diff
@@ -0,0 +1,229 @@
...
+    )
+    loop = asyncio.get_running_loop()
+    rows = await loop.run_in_executor(
+        None,
+        lambda: db.execute(
+            stmt,
+            {"vec": vec_str, "kb_id": str(kb_id), "top_k": top_k},
+        ).fetchall(),
+    )
+
```
codewithwaleed, line 98: **SQLAlchemy `Session` is not thread-safe / not safe for concurrent use.** `retrieve_kb_context_for_turn` calls `_query_single_kb` for every `kb_id` via `asyncio.gather(...)`, and each one does `loop.run_in_executor(None, lambda: db.execute(stmt, ...))` on the **same** `db: Session` instance. With more than one KB attached to a call flow, multiple threads will execute against the same `Session`/connection concurrently, which can raise `InvalidRequestError` ("This session is provisioning a new connection; concurrent operations are not permitted") or, worse, silently interleave results.

**Solution:** either run the per-KB queries sequentially (the latency budget is already small and pgvector queries are fast), or give each parallel query its own short-lived session/connection, e.g.:
```python
from app.db.session import SessionLocal

async def _query_single_kb(kb_id, vec_str, top_k):
    def _run():
        with SessionLocal() as session:
            return session.execute(stmt, {...}).fetchall()
    return await loop.run_in_executor(None, _run)
```
so each task in `asyncio.gather` operates on its own `Session`/connection.

codewithwaleed (changes requested):
Reviewed BE1-S7 real-time KB retrieval layer. The retrieval/caching logic in `kb_retrieval_service.py` and the latency-budgeted wiring in `conversation_orchestrator.py` are solid (fail-open, 0.45s timeout, parallel per-KB queries, Redis caching). However there's a critical migration bug around the `usagerecord` table that will destroy billing-audit data, plus a SQLAlchemy Session thread-safety issue in the parallel KB query path.

PR #97, opened 2026-06-15
feat(callback): Smart Callback Scheduler — timezone-aware outbound retry with APScheduler

Aliict:
## Summary

Implements the **Smart Callback Scheduler** — an automated retry system for outbound calls that end as `no_answer` or `busy`. Retries are timezone-aware (business hours enforced), fully traceable via `parent_call_id`, and survive pod restarts via a PostgreSQL-backed APScheduler job store.

### Database
- **Alembic migration** (`20260615_callback`): adds 4 columns to `agent` (`smart_callback_enabled`, `max_callback_attempts`, `callback_gap_schedule JSONB`, `callback_timezone`), adds `parent_call_id` self-FK to `callsession`, creates new `callbackschedule` table with status CHECK constraint and composite index on `(status, scheduled_at)`

### New files
| File | Purpose |
|---|---|
| `app/models/callback_schedule.py` | `CallbackSchedule` ORM model — table name `callbackschedule` (base class convention) |
| `app/schemas/callback_scheduler.py` | Pydantic v2 schemas: `CallbackConfigUpdate` (with IANA timezone validation), `CallbackStatusResponse`, `CallbackHistoryItem` |
| `app/services/callback_scheduler_service.py` | All business logic: config CRUD, trigger detection, business-hours enforcement, real outbound dispatch, attempt chaining, exhaustion |
| `app/core/scheduler.py` | APScheduler singleton with `SQLAlchemyJobStore` (PostgreSQL), `IntervalTrigger(seconds=30)`, `coalesce=True`, `max_instances=1` |
| `app/api/v2/routers/callback_scheduler.py` | v2 endpoints — two sub-routers at `/agents` and `/calls` prefixes |
| `tests/api/test_callback_scheduler.py` | 18 pytest cases covering all acceptance criteria |

### Modified files
- `app/models/agent.py` — 4 new smart-callback columns + `callback_schedules` relationship
- `app/models/call_session.py` — `parent_call_id` self-FK + relationships
- `app/services/call_session_service.py` — hooks `maybe_schedule_callback` into `update_call_session_status` for `no_answer`/`busy` transitions
- `app/api/v2/api.py` — registers callback sub-routers
- `app/main.py` — starts/stops APScheduler in the lifespan context
- `app/db/base.py` — adds `CallbackSchedule` import for Alembic autogenerate
- `requirements.txt` — adds `APScheduler==3.11.0`

## API Endpoints (v2)

```
PUT  /api/v2/agents/{agent_id}/callback-config    — set enabled, max_attempts, gap_schedule, timezone
GET  /api/v2/agents/{agent_id}/callback-status    — live counters (pending_retries, next_scheduled_at)
GET  /api/v2/calls/{call_id}/callback-history     — ordered list of all retry attempts
```

## How the retry chain works

```
Call ends as no_answer / busy
        │
        ▼
maybe_schedule_callback()
  ├─ agent.smart_callback_enabled? No → skip
  └─ Yes → insert CallbackSchedule(attempt=1, scheduled_at=now+gap_schedule[0])
                    │
                    ▼  (APScheduler fires every 30 s)
        process_pending_callbacks()
          ├─ scheduled_at <= now?  No → wait
          └─ Yes → _dispatch_and_advance()
                    ├─ Within business hours?  No → reschedule to next open window
                    └─ Yes → initiate_call() [real Twilio/LiveKit dispatch]
                              ├─ attempt < max_attempts → insert next CallbackSchedule
                              └─ attempt == max_attempts → status = 'exhausted', stop
```

## Key design decisions

- **APScheduler PostgreSQL job store** — `apscheduler_jobs` table auto-managed; survives pod restarts
- **SKIP LOCKED** — concurrent pods never double-dispatch the same record
- **Fresh event loop per dispatch** — APScheduler thread-pool threads have no running loop; `asyncio.new_event_loop()` is safe and avoids cross-loop contamination
- **Same auth path as batch calls** — uses `N8N_WEBHOOK_SECRET` webhook header so `initiate_call` requires no User object
- **`parent_call_id` traceability** — post-updated on the new `CallSession` after dispatch; full chain navigable via FK
- **Business hours fallback** — 14-day forward-walk in 1-hour increments; falls back to original time if no open window found (prevents chain stall)
- **Non-blocking hook** — callback scheduling failure in `update_call_session_status` logs a warning and never breaks the call status update

## Test plan

- [x] Run migration: `alembic upgrade head`
- [x] Install dependency: `pip install APScheduler==3.11.0`
- [x] Set `N8N_WEBHOOK_SECRET` in `.env`
- [x] Run tests: `pytest tests/api/test_callback_scheduler.py -v`
- [x] Create an agent, call `PUT /api/v2/agents/{id}/callback-config` with a valid gap schedule
- [x] Trigger an outbound call that results in `no_answer` → confirm `callbackschedule` row created
- [x] Wait 30 s → confirm APScheduler dispatches the retry and creates a new `callsession` with `parent_call_id` set
- [x] Call `GET /api/v2/calls/{call_id}/callback-history` → verify attempt history
- [x] Let chain run to `max_attempts` → confirm final record is `exhausted`

### DIFF
--- a/app/services/callback_scheduler_service.py
+++ b/app/services/callback_scheduler_service.py
@@ -222,17 +222,30 @@ class CallbackSchedulerService:
         """
         Poll for due callbacks and dispatch each one.
         Designed to be called by the APScheduler IntervalTrigger job.
+
+        Each iteration selects a single row with FOR UPDATE SKIP LOCKED so the
+        lock is scoped to exactly one transaction. Committing (or rolling back)
+        inside _dispatch_and_advance releases only that row's lock, which means
+        other workers/pods can safely claim any remaining rows without risk of
+        double-dispatch.
         """
         now = datetime.utcnow().replace(tzinfo=ZoneInfo("UTC"))
 
-        due_rows = db.execute(
-            select(CallbackSchedule).where(
-                CallbackSchedule.status == "pending",
-                CallbackSchedule.scheduled_at <= now,
-            ).with_for_update(skip_locked=True)
-        ).scalars().all()
+        while True:
+            schedule = db.execute(
+                select(CallbackSchedule)
+                .where(
+                    CallbackSchedule.status == "pending",
+                    CallbackSchedule.scheduled_at <= now,
+                )
+                .order_by(CallbackSchedule.scheduled_at)
+                .with_for_update(skip_locked=True)
+                .limit(1)
+            ).scalar_one_or_none()
 
-        for schedule in due_rows:
+            if schedule is None:
+                break
+
             try:
                 self._dispatch_and_advance(db, schedule)
             except Exception as exc:  # noqa: BLE001
@@ -348,12 +361,23 @@ class CallbackSchedulerService:
         """
         Dispatch the outbound callback call via the existing initiate_call path.
 
-        Runs in an APScheduler thread-pool thread (no event loop), so we create
-        a fresh event loop for the async call and a dedicated DB session so the
-        async path can commit without interfering with the caller's session.
+        Session boundaries
+        ------------------
+        dispatch_db is intentionally separate from the poller session (db).
+        initiate_call commits its own work (new CallSession, Twilio request)
+        inside dispatch_db. A db.rollback() in the poller NEVER rolls back
+        dispatch_db — the call delivery is guaranteed to persist regardless of
+        any subsequent bookkeeping failure in the poller session.
 
-        After dispatch the new CallSession.parent_call_id is set to
-        schedule.original_call_id for full call-chain traceability.
+        Failure semantics
+        -----------------
+        - initiate_call failure   → real failure; raise so the poller can
+                                    rollback and leave the row in 'pending'.
+        - parent_call_id update failure → non-critical bookkeeping; logged
+                                          but does NOT fail the dispatch.
+
+        Runs in an APScheduler thread-pool thread (no event loop), so we spin
+        up a fresh event loop for the async call.
         """
         import asyncio
         import json as _json
@@ -413,7 +437,9 @@ class CallbackSchedulerService:
 
         fake_request = _Request(scope, receive=_receive)
 
-        # ── Run async initiate_call in a fresh event loop ─────────────────────
+        # ── Phase 1: call initiation (dispatch_db, independent of poller db) ────
+        # dispatch_db owns its own transaction. Commits inside initiate_call are
+        # durable regardless of what happens to the poller session afterwards.
         # APScheduler thread has no running loop, so new_event_loop() is safe.
         dispatch_db = SessionLocal()
         result = None
@@ -433,9 +459,10 @@ class CallbackSchedulerService:
                 loop.close()
                 asyncio.set_event_loop(None)
         finally:
+            # Always close dispatch_db; its commits are already flushed to PG.
             dispatch_db.close()
 
-        # ── Extract the new session id and set parent_call_id ─────────────────
+        # ── Extract new session id — failure here = real initiation failure ────
         if isinstance(result, _JSONResponse):
             body = _json.loads(result.body)
             new_call_id_str = (body.get("data") or {}).get("callId")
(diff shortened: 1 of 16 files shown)


app/services/callback_scheduler_service.py
```diff
@@ -0,0 +1,557 @@
...
+        Poll for due callbacks and dispatch each one.
+        Designed to be called by the APScheduler IntervalTrigger job.
+        """
+        now = datetime.utcnow().replace(tzinfo=ZoneInfo("UTC"))
+
+        due_rows = db.execute(
+            select(CallbackSchedule).where(
+                CallbackSchedule.status == "pending",
+                CallbackSchedule.scheduled_at <= now,
+            ).with_for_update(skip_locked=True)
```
codewithwaleed: **`SKIP LOCKED` protection is lost after the first row's commit.** `process_pending_callbacks` does a single `SELECT ... FOR UPDATE SKIP LOCKED` that locks *all* `due_rows` for the duration of the current transaction. But `_dispatch_and_advance(db, schedule)` calls `db.commit()` for the first row — which ends that transaction and **releases the row locks on every other row in `due_rows`**, not just the one just processed.

If a second pod's poller runs concurrently (or the loop is slow and another 30s tick fires), it can now `SELECT ... FOR UPDATE SKIP LOCKED` and pick up the *same* remaining rows from `due_rows` that this loop hasn't gotten to yet — both pods will dispatch the same callback.

**Solution:** lock and dispatch one row per transaction, e.g.:
```python
def process_pending_callbacks(self, db: Session) -> None:
    now = datetime.utcnow().replace(tzinfo=ZoneInfo("UTC"))
    while True:
        schedule = db.execute(
            select(CallbackSchedule)
            .where(CallbackSchedule.status == "pending", CallbackSchedule.scheduled_at <= now)
            .order_by(CallbackSchedule.scheduled_at)
            .with_for_update(skip_locked=True)
            .limit(1)
        ).scalar_one_or_none()
        if schedule is None:
            break
        try:
            self._dispatch_and_advance(db, schedule)
        except Exception as exc:
            logger.error("callback_dispatch_error schedule_id=%s: %s", schedule.id, exc, exc_info=True)
            db.rollback()
```
Each iteration's `SELECT ... FOR UPDATE` + `_dispatch_and_advance`'s commit now form one transaction, so the lock is held until that specific row is finalized.

app/services/callback_scheduler_service.py
```diff
@@ -0,0 +1,557 @@
...
+            dispatch_db.close()
+
+        # ── Extract the new session id and set parent_call_id ─────────────────
+        if isinstance(result, _JSONResponse):
+            body = _json.loads(result.body)
+            new_call_id_str = (body.get("data") or {}).get("callId")
+            if not new_call_id_str:
+                raise RuntimeError(
+                    f"initiate_call failed for callback (attempt {schedule.attempt_number}): "
+                    f"{body.get('message', 'unknown error')}"
```
codewithwaleed, line 472: Minor: `dispatch_db = SessionLocal()` is opened for `initiate_call` but only `db` (the poller's session) is used to read `original` and write `parent_call_id` back. If `initiate_call` raises partway through after having committed some state via `dispatch_db`, that commit isn't rolled back when the outer exception handler in `process_pending_callbacks` calls `db.rollback()` (that only rolls back `db`, not `dispatch_db`, which is already closed). Worth a comment noting this is intentional (the dispatched call should persist even if the bookkeeping update fails), or wrap the `parent_call_id` update in its own try/except so a failure there doesn't get logged as a full dispatch failure for an outbound call that actually went out.

codewithwaleed (changes requested):
Reviewed the Smart Callback Scheduler. The retry-chain model, business-hours fallback, and APScheduler/SQLAlchemy job store setup are well designed. One concurrency issue undermines the stated "concurrent pods never double-dispatch" guarantee, plus a smaller robustness note on the dispatch path.

PR #98, opened 2026-06-15
feat(integrations): Make.com and n8n workflow trigger endpoints

Aliict:
## Summary

- Implements workspace-scoped outbound call triggers for **Make.com** and **n8n** so external automation workflows can dispatch calls without needing JWT auth
- Both integrations share the same internal `voice_call_service.initiate_call` dispatch — no duplicated call logic
- Adds `workspace_settings` JSONB column to the `tenant` table (migration included)

## Endpoints

| Method | Path | Auth | Description |
|---|---|---|---|
| `POST` | `/api/v1/workspace/settings/make-secret` | JWT or API key | Generate / rotate the Make.com workspace secret |
| `POST` | `/api/v1/integrations/make/trigger` | `X-Make-Secret` header | Trigger outbound call from Make.com scenario |
| `POST` | `/api/v1/integrations/n8n/trigger` | `X-N8N-Webhook-Secret` header | Trigger outbound call from n8n HTTP Request node |
| `GET` | `/api/v1/integrations` | JWT or API key | List integrations with `connected`, `webhook_url`, `last_triggered_at` |

## Key Design Decisions

- **Make.com secret** is workspace-scoped: 64-char hex generated via `secrets.token_hex(32)`, stored in `tenant.workspace_settings.make_secret`. Rotating calls the same endpoint again.
- **Internal dispatch** uses the established `_build_internal_request` pattern (same as `batch_call_worker_service`) — builds a minimal Starlette Request with `X-N8N-Webhook-Secret` so `initiate_call` resolves the webhook auth path.
- **Rate limit**: 10 integration-triggered calls per minute per workspace via Redis sliding window. Returns `429` with `retry_after` timestamp.
- **n8n connected** status is derived from whether `N8N_WEBHOOK_SECRET` env var is set (per the existing system contract).

## Test Plan

- [x] Make.com happy path — valid secret dispatches call, returns `{call_id, status}`
- [x] Make.com invalid secret → `403 {message: "Invalid secret", code: "unauthorized"}`
- [x] Make.com missing secret → `403`
- [x] Make.com unknown agent → `404`
- [x] Make.com rate limit → `429` with `retry_after`
- [x] n8n happy path — returns `{success: true, data: {call_id, status}}`
- [x] n8n invalid secret → `403`
- [x] n8n missing `tenant_id` → `400`
- [x] n8n rate limit → `429`
- [x] Integration list — `connected` reflects secret presence
- [x] `last_triggered_at` round-trip (record + read back as datetime)
- [x] Secret generation format (64-char hex, unique on each call)
- [x] Secret rotation replaces previous value
- [x] Rate limit fail-open when Redis unavailable
- [x] Rate limit enforced via Redis pipeline

**19 tests, 0 regressions** (pre-existing failures confirmed on base branch before changes).

### DIFF
--- a/app/routers/integrations.py
+++ b/app/routers/integrations.py
@@ -1,0 +1,309 @@
+"""
+Integration trigger endpoints for Make.com and n8n.
+
+Both endpoints share the same internal call dispatch (voice_call_service.initiate_call).
+Auth is handled per-integration:
+  - Make.com: X-Make-Secret header validated against workspace_settings.make_secret
+  - n8n:      X-N8N-Webhook-Secret header validated against settings.N8N_WEBHOOK_SECRET
+
+Rate limit: 10 integration-triggered calls per minute per workspace.
+"""
+from __future__ import annotations
+
+import uuid
+from datetime import datetime, timezone
+from typing import Optional
+
+from fastapi import APIRouter, Depends, Header, HTTPException, Request, status
+from fastapi.responses import JSONResponse
+from sqlalchemy.orm import Session
+
+from app.api.deps import get_db, require_tenant
+from app.core.config import settings
+from app.core.logger import logger
+from app.schemas.integration import (
+    IntegrationItem,
+    IntegrationListResponse,
+    MakeTriggerRequest,
+    MakeTriggerResponse,
+    N8nTriggerResponse,
+)
+from app.schemas.twilio import CallInitiateRequest
+from app.services.integration_service import (
+    check_integration_rate_limit,
+    get_last_triggered_at,
+    get_make_secret,
+    record_last_triggered,
+    resolve_tenant_by_agent,
+)
+from app.services.voice_call_service import initiate_call as initiate_call_service
+from app.utils.n8n_webhook_verification import verify_n8n_webhook_secret_async
+
+router = APIRouter()
+
+
+def _build_internal_request(webhook_secret: str) -> Request:
+    """
+    Build a minimal Starlette Request for internal outbound call dispatch.
+
+    Mirrors the pattern in batch_call_worker_service._build_fake_request so that
+    verify_n8n_webhook_secret_async resolves to True and initiate_call uses the
+    tenant_id from the body rather than requiring a JWT user.
+    """
+    from starlette.types import Scope
+
+    scope: Scope = {
+        "type": "http",
+        "method": "POST",
+        "path": "/internal/integration",
+        "query_string": b"",
+        "headers": [
+            (b"x-n8n-webhook-secret", webhook_secret.encode("latin-1")),
+            (b"content-type", b"application/json"),
+        ],
+        "state": {},
+    }
+
+    async def _receive():
+        return {"type": "http.request", "body": b"", "more_body": False}
+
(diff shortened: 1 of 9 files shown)


app/routers/integrations.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+
+    # 2. Resolve tenant for rate limiting
+    if not body.tenant_id:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="tenant_id is required in request body",
+        )
+    try:
+        workspace_id = uuid.UUID(body.tenant_id)
```
codewithwaleed, line 221: **Cross-tenant call dispatch via a single global secret.** `/n8n/trigger` is authorized solely by `verify_n8n_webhook_secret_async(request)`, which checks the **single global** `settings.N8N_WEBHOOK_SECRET` — the same value for every tenant in the system. Authorization for *which tenant* the call is placed under then comes entirely from `body.tenant_id`, supplied by the caller with no ownership check.

Any party that obtains this one global secret (e.g. a tenant configures it in their own n8n instance, or it leaks) can dispatch outbound calls billed to/under **any other tenant** simply by setting `tenant_id` to that tenant's UUID in the request body — there's no verification that the caller is associated with `workspace_id`.

Contrast with `/make/trigger`, which correctly uses a per-workspace secret (`workspace_settings.make_secret`) so a Make.com scenario can only act on its own tenant.

**Solution:** give n8n the same per-workspace secret model as Make (e.g. `workspace_settings.n8n_secret`, generated/rotated via an analogous endpoint, validated via an `X-N8N-Webhook-Secret` header that's compared against the *resolved tenant's* stored secret rather than a single global env var). If the global `N8N_WEBHOOK_SECRET` must be kept for backward compatibility with the existing internal dispatch pattern, at minimum add a secondary check that the global-secret caller is only allowed to use it for dispatch (internal/batch use), and require the new public-facing `/n8n/trigger` endpoint to authenticate with a tenant-scoped credential.

app/routers/integrations.py
```diff
@@ -0,0 +1,309 @@
...
+    agent, tenant = resolve_tenant_by_agent(db, body.agent_id)
+    if agent is None or tenant is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail="Agent not found",
+        )
+
+    # 2. Validate X-Make-Secret against workspace_settings
+    stored_secret = get_make_secret(tenant)
+    if not stored_secret or not x_make_secret or x_make_secret != stored_secret:
```
codewithwaleed, line 129: Use a constant-time comparison for the secret check to avoid timing side-channels: `import hmac; hmac.compare_digest(x_make_secret, stored_secret)` instead of `!=`.

codewithwaleed (changes requested):
Reviewed the Make.com/n8n integration trigger endpoints. The Make.com flow (per-workspace secret in `workspace_settings`) is correctly tenant-scoped. The n8n flow has a serious cross-tenant authorization issue, plus a timing-safe comparison nit on the Make secret check.

PR #104, opened 2026-06-17
Feat/db schema v3 updates

ahbilal220:
This PR introduces the foundational database schema and models for the V3 Hierarchical Multi-Tenant Platform, enabling Agency/Sub-Account structures, workspace branding, custom pricing configurations, and granular usage tracking.

### Key Changes
- **Hierarchical Workspaces (Multi-Tenancy):**
  - Updated the `tenant` model to support an adjacency list structure by adding `parent_workspace_id` and `workspace_type` (agency, sub_account, standalone). This allows Sub-Accounts to be linked to a top-level Agency workspace.
- **Workspace Branding & Pricing:**
  - Added the `branding_configs` table (1:1 with `tenant`) to support custom logos, primary colors, and display names.
  - Added the `pricing_configs` table (1:1 with `tenant`) to manage custom `per_minute_rate` overrides and `markup_percent`.
- **Usage Tracking Ledger:**
  - Updated `usagerecord` table linked to workspaces and call sessions, tracking `billable_minutes` (call duration in seconds / 60) for precise billing calculations.
- **RBAC Refactor & Cleanup:**
  - Reverted an experimental approach of adding workspace/user foreign keys directly to the `Role` model. We are strictly adhering to the existing `user_tenant_association` mapping table, keeping the `Role` table as a clean, global lookup definition. This avoids structural duplication and keeps the design consistent.
- **Local Dev & Seeding:**
  - Updated `scripts/seed_dev_workspace.py` to automatically seed one Parent Agency workspace, one Linked Sub-Account, and assign the local admin user to both environments correctly.
  - Replaced missing schema dependencies for fresh setups with an `init_fresh_db.py` script that safely creates all tables (bypassing missing `pgvector` extensions dynamically) and stamps Alembic to `head`.
- **Documentation & Linting:**
  - Updated the Schema V3 documentation and Mermaid ER diagrams (`docs/db/schema-v3.md`) to perfectly reflect the final, simplified structural state.
  - Fixed widespread type-checker false-positives across all models by adding a `**kwargs` `__init__` stub to the SQLAlchemy `Base` class.

### Technical Notes
- All new tables follow the established V1/V2 conventions (UUID primary keys, `created_at`/`deleted_at` soft deletes).
- Alembic migrations have been tested for both clean upgrades (`alembic upgrade head`) and downgrades (`alembic downgrade -1`).
- The `parent_workspace_id` being `NULL` denotes a top-level workspace (Agency or Standalone).


app/models/role.py
Aliict, line 1: We should stick with the existing user_tenant_association table for handling user-to-workspace role mapping. The current changes introduce unnecessary complexity by adding user_id and workspace_id to the role table. These new columns are not required and create duplication in responsibility, so they should be removed to keep the design clean and consistent.

PR #105, opened 2026-06-17
Feat/workspace branding and pricing config api

ahbilal220:
## 🚀 Description

This PR implements the P1 endpoints for **Workspace Branding**, **Pricing Configuration**, and **Usage Tracking**. These endpoints allow agency owners to fully customize their workspace appearance (used by the frontend to theme the UI per agency) and set per-client minute markup rates dynamically.

## 📝 Key Changes

- **Pydantic Schemas (`app/schemas/workspace.py`)**: 
  - Added strict validation for configurations: `primary_colour` enforces hex codes (`^#[0-9A-Fa-f]{6}$`), `logo_url` enforces HTTPS, and `markup_percent` is constrained between 0-500.
  - Implemented `PricingConfigOut` to dynamically compute `effective_client_rate` on the fly.
  - Implemented `WorkspaceUsageOut` to compute `minutes_used_this_cycle` and `overage_cost`.
- **v2 Endpoints (`app/api/api_v1/endpoints/workspace.py`)**: 
  - Added a new `v2_router` inside the existing workspace module to prevent fragmentation, while properly routing it to the `/api/v2/workspace` namespace.
  - Built `GET` and `PUT` upsert endpoints for `/branding` and `/pricing` with PostgreSQL `INSERT ... ON CONFLICT DO UPDATE`.
  - Built the `GET /usage` endpoint, utilizing DB aggregation (`func.sum` and `func.date_trunc('month', now())`) to calculate MTD usage across sub-accounts dynamically.
  - Added strict defensive casting to `Decimal` objects across all financial math blocks to avoid python float/Decimal mismatches.
- **RBAC**: Handled correctly. All 5 endpoints are protected by `require_admin` to ensure only administrators or owners can touch configuration/pricing states.
- **Testing (`tests/api/v2/test_workspace_config.py`)**: Authored comprehensive integration tests verifying JSON validation (422s), upsert operations, calculations, and RBAC rejection (403).
- **Docs**: Re-ran the export script to update `docs/api/openapi.yaml` with the new schemas and definitions.

## 🧪 Testing Instructions
1. Check out the branch.
2. Spin up your local dev environment.
3. Validate tests passing via `venv/Scripts/pytest tests/api/v2/test_workspace_config.py` (or through the docker test suite).
4. Go to Swagger UI `/docs` and verify the `/api/v2/workspace` routes map correctly.

### DIFF
--- a/app/api/api_v1/endpoints/workspace.py
+++ b/app/api/api_v1/endpoints/workspace.py
@@ -13,15 +13,12 @@ from pydantic import ValidationError
 from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 from sqlalchemy.orm import Session
 
-from app.api.deps import get_db, get_workspace_api_key, require_tenant, require_admin
+from app.api.deps import get_db, get_workspace_api_key, require_tenant
 from app.core.request_auth import get_workspace_from_request
 from app.core.config import settings
 from app.core.logger import logger
 from app.core.workspace import Workspace
 from app.models.tenant import Tenant
-from app.models.branding_configs import BrandingConfig
-from app.models.pricing_configs import PricingConfig
-from app.models.usage_record import UsageRecord
 from app.repositories.workspace_repository import WorkspaceRepository
 from app.schemas.base import SuccessResponse
 from app.schemas.integration import MakeSecretResponse, N8nSecretResponse
@@ -30,11 +27,6 @@ from app.schemas.workspace import (
     WorkspaceCreatedOut,
     WorkspaceOut,
     WorkspaceUpdateName,
-    BrandingConfigUpsert,
-    BrandingConfigOut,
-    PricingConfigUpsert,
-    PricingConfigOut,
-    WorkspaceUsageOut,
 )
 from app.services.integration_service import (
     generate_make_secret,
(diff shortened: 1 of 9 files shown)


app/api/api_v1/endpoints/workspace.py
```diff
@@ -380,3 +388,134 @@ def soft_delete_workspace(
...
+    db: Session = Depends(get_db),
+):
+    """Upsert the pricing configuration for the workspace."""
+    from decimal import Decimal
+    from sqlalchemy.dialects.postgresql import insert
+    stmt = insert(PricingConfig).values(
+        workspace_id=user.current_tenant_id,
+        per_minute_rate=payload.per_minute_rate,
+        markup_percent=payload.markup_percent,
+    )
```
codewithwaleed: **`minutes_included` is hardcoded to `0`, making every minute an overage — this will overcharge every workspace.** There are no plan-tier lookups here, so any customer using the voice agent will see 100% of their usage billed as overage from day one.

**Solution:** pull the included minutes from the workspace's active subscription/plan:
```python
from app.models.subscription import Subscription  # or your billing model

sub = db.query(Subscription).filter(
    Subscription.workspace_id == user.current_tenant_id,
    Subscription.status == "active",
).first()
minutes_included = Decimal(str(sub.minutes_included)) if sub else Decimal("0")
```
If billing plans are not yet implemented, at minimum return `minutes_included=None` and document the field as "coming soon" rather than silently lying with `0`.

app/models/role.py
```diff
@@ -14,7 +14,7 @@ class Role(Base):
     role = Column(String, nullable=True)
     deleted_at = Column(DateTime(timezone=True), nullable=True)
 
-    tenant = relationship("Tenant", back_populates="rbac_roles")
+    # tenant = relationship("Tenant", back_populates="rbac_roles")
```
codewithwaleed: **Commented-out `back_populates` breaks SQLAlchemy bidirectional relationships at runtime.** Any code that still accesses `tenant.rbac_roles` (or `role.tenant`) will get an `AttributeError` because `Tenant.rbac_roles` is also commented out. Commenting is worse than removing — it leaves a broken stub.

**Solution:** either delete both sides of the relationship entirely from `Role` and `Tenant`, or restore them. If the relationship is genuinely being dropped, remove it from both models and remove `cascade="all, delete-orphan"` from `Tenant.rbac_roles` so SQLAlchemy's mapper doesn't try to resolve a missing back-populates target at startup.

codewithwaleed (changes requested):
Reviewed workspace branding, pricing, and usage config endpoints. The upsert pattern with `ON CONFLICT DO UPDATE` is clean, and the `require_admin` guard is correctly applied everywhere. Two issues need fixing before merge: a hardcoded zero for included minutes that will overcharge all customers, and commented-out ORM relationships that will cause runtime AttributeErrors.

app/api/api_v1/endpoints/workspace.py
```diff
@@ -380,3 +388,134 @@ def soft_delete_workspace(
         old_value={"name": old_name},
     )
     return Response(status_code=status.HTTP_204_NO_CONTENT)
+
+
+v2_router = APIRouter()
+
+@v2_router.get("/branding", response_model=BrandingConfigOut)
```
Aliict: Anti-pattern: V2 endpoints should not be placed inside V1 workspace file.

This breaks versioning separation and code structure clarity.

Either keep these endpoints under V1 if they are truly V1, or properly move them into the V2 router (app/api/v2/routers/workspace.py).

app/models/role.py
```diff
@@ -13,9 +13,7 @@ class Role(Base):
     created_at = Column(DateTime(timezone=True), server_default=func.now(), nullable=False)
     role = Column(String, nullable=True)
     deleted_at = Column(DateTime(timezone=True), nullable=True)
-
-    tenant = relationship("Tenant", back_populates="rbac_roles")
-
+    
     __table_args__ = (
```
Aliict: This CheckConstraint and the 'role' column are unused and inconsistent with the actual application roles.

 The system uses the 'name' column for authorization (owner, admin, member, config, readonly),
 while this constraint defines a different and outdated role set.

 It should be removed to avoid confusion and schema mismatch.

PR #106, opened 2026-06-18
feat(gdpr): data export & right-to-erasure account deletion

Aliict:
## Summary
- Adds `POST/GET /api/v2/workspace/data-export` — async ARQ-backed export of all workspace data (metadata, calls, transcripts, KB chunks, audit events, batch records) streamed into a ZIP, uploaded to GCS, with a 24h signed URL emailed to the workspace admin.
- Adds `DELETE /api/v2/workspace/account` — exact-phrase-gated (`DELETE MY ACCOUNT`, case-sensitive), irreversible PII wipe + GCS recording/KB-embedding deletion + audit-actor anonymization + workspace soft-delete, via a single raw-SQL transaction.
- Fixes three Postgres schema issues discovered while implementing the literal ticket spec against real Postgres (not just SQLite unit tests) — see "Schema changes" below.

## Schema changes (already applied to the shared dev DB)
1. **`auditlog` append-only enforcement**: replaced the unconditional `no_update_audit` RULE (added in #103) with a `BEFORE UPDATE` trigger gated on a new `app.bypass_audit_update` GUC. When set, the trigger ignores whatever the UPDATE statement requests and force-applies only `user_id=NULL, actor_api_key_prefix='[DELETED]'` — every other column reverts to OLD. This lets erasure anonymize actor fields while keeping the table append-only for every other write path, with the bypass's blast radius capped to exactly that one transformation.
2. **`user.email` uniqueness**: was a global `UNIQUE` constraint. Erasure anonymizes every deleted user's email to the same literal placeholder, so a second deletion anywhere would have violated it. Replaced with a partial unique index (`WHERE deleted_at IS NULL`), mirroring the existing `uq_tenant_name_active` pattern.
3. **`auditlog.actor_api_key_prefix`** was `VARCHAR(8)` (sized for real 8-char API key prefixes) — one character short of the 9-char `'[DELETED]'` placeholder. Widened to `VARCHAR(16)`.

Migrations: `20260618_audit_update_bypass`, `20260618_data_export_job`, `20260618_user_email_partial_unique`, `20260618_audit_actor_prefix_widen`. Already run against dev (Neon) and verified column-by-column (types, indexes, triggers) post-upgrade.

## Design notes
- A `User` can belong to multiple workspaces. Only users whose *sole* membership is the workspace being erased get their identity wiped; users shared with other active workspaces just lose membership in this one workspace — their identity elsewhere is untouched.
- Both endpoints require the existing `require_admin` dependency, matching the established convention for other workspace-level mutations (HIPAA kms-key/status).
- ZIP is built via streaming writes (`ZipFile.open(..., "w")`, DB rows fetched with `yield_per`) so the export is never fully buffered in memory.

## Test plan
- [x] Unit tests (SQLite, mocked GCS/email/ARQ): 24 tests across endpoint contracts (`tests/api/v2/test_data_export.py`, `tests/api/v2/test_account_deletion.py`), the zip builder's real DB→ZIP content (`tests/services/test_data_export_zip_builder.py`), and the full export orchestration + ARQ task wrapper (`tests/services/test_data_export_service.py`)
- [x] Integration tests against real Postgres (`tests/integration/test_gdpr_account_deletion_postgres.py`, gated by `TEST_DATABASE_URL`): trigger blocks default updates, bypass GUC anonymizes only the intended columns regardless of what's requested, full PII wipe verified end-to-end (multi-tenant user carve-out, phone/call redaction, KB chunk deletion, audit retention + anonymization, workspace soft-delete), wrong-phrase 400 over a real JWT session
- [x] Confirmed via diff against `dev` that pre-existing unrelated test failures (~190 in the full suite, plus a pre-existing Postgres-test-harness JWT/Redis flakiness) are unaffected by this change
- [x] Migrations applied to dev DB and schema verified directly (column types, indexes, triggers)

### DIFF
--- a/app/api/v2/routers/workspace.py
+++ b/app/api/v2/routers/workspace.py
@@ -1,0 +1,187 @@
+"""
+v2 GDPR data subject rights router.
+
+Endpoints (admin role required for all):
+  POST   /api/v2/workspace/data-export             — trigger async export, returns 202 {job_id}
+  GET    /api/v2/workspace/data-export/{job_id}     — export job status + signed download URL
+  DELETE /api/v2/workspace/account                  — irreversible hard delete + PII wipe
+"""
+from __future__ import annotations
+
+import uuid
+from typing import Optional
+
+from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
+from pydantic import BaseModel
+from sqlalchemy.orm import Session
+
+from app.api.deps import get_db, require_admin
+from app.core.logger import logger
+from app.models.user import User
+from app.services.account_deletion_service import delete_workspace_account
+from app.services.audit_service import log_audit_event
+from app.services.data_export_service import create_export_job, get_export_job
+
+router = APIRouter(prefix="/workspace", tags=["workspace-gdpr"])
+
+_DELETE_CONFIRMATION_PHRASE = "DELETE MY ACCOUNT"
+
+
+# ── Schemas ───────────────────────────────────────────────────────────────────
+
+
+class DataExportTriggerOut(BaseModel):
+    job_id: uuid.UUID
+
+
+class DataExportStatusOut(BaseModel):
+    status: str
+    download_url: Optional[str] = None
+
+
+class AccountDeletionRequest(BaseModel):
+    confirmation: str
+
+
+# ── ARQ enqueue helper ────────────────────────────────────────────────────────
+
+
+async def _enqueue_data_export_job(export_job_id: str) -> None:
+    """
+    Push a run_data_export_job task into ARQ. Falls back to a temporary
+    per-call pool if the shared pool was not initialised, same as the
+    batch-calls enqueue helper. Never raises — caller can still return 202
+    and the job stays 'processing' until an operator notices and re-runs it.
+    """
+    try:
+        from app.utils.arq_pool import get_arq_pool
+
+        pool = get_arq_pool()
+        _owns_pool = False
+
+        if pool is None:
+            import arq  # type: ignore
+            from app.core.config import settings as cfg
+
+            redis_settings = arq.connections.RedisSettings.from_dsn(cfg.REDIS_URL)
+            pool = await arq.create_pool(redis_settings)
+            _owns_pool = True
+
+        try:
+            await pool.enqueue_job("run_data_export_job", export_job_id)
+            logger.info("DataExportJob %s enqueued in ARQ", export_job_id)
+        finally:
+            if _owns_pool:
+                await pool.aclose()
+
+    except Exception as exc:
+        logger.warning(
+            "ARQ enqueue failed for data export %s: %s",
+            export_job_id,
+            exc,
+        )
+
+
+# ── POST /workspace/data-export ─────────────────────────────────────────────
+
+
+@router.post(
+    "/data-export",
+    response_model=DataExportTriggerOut,
+    status_code=status.HTTP_202_ACCEPTED,
+)
+async def trigger_data_export(
+    request: Request,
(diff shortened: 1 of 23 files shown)


app/api/v2/routers/workspace.py
```diff
@@ -0,0 +1,187 @@
...
+
+# ── DELETE /workspace/account ─────────────────────────────────────────────────
+
+
+@router.delete(
+    "/account",
+    status_code=status.HTTP_204_NO_CONTENT,
+    response_class=Response,
+)
+def delete_account(
```
codewithwaleed, line 153: **HTTP `DELETE` with a request body is unreliable** — many reverse proxies (nginx, AWS ALB) and some HTTP clients drop the body on DELETE requests, meaning `body.confirmation` arrives as empty/null and the endpoint always returns 400, or worse: some proxies pass an empty body through and the confirmation check is bypassed.

**Solution:** change the method to `POST` with an explicit action path:
```python
@router.post(
    "/account/delete",
    status_code=status.HTTP_204_NO_CONTENT,
    response_class=Response,
)
def delete_account(body: AccountDeletionRequest, ...) -> Response:
    ...
```
or move the confirmation phrase to a custom header (`X-Confirm-Delete: DELETE MY ACCOUNT`) on the DELETE route — headers are always forwarded.

app/services/account_deletion_service.py
```diff
@@ -0,0 +1,136 @@
...
+
+        db.execute(
+            text(
+                "UPDATE callsession SET from_number = :anon_phone, to_number = :anon_phone "
+                "WHERE tenant_id = :workspace_id"
+            ),
+            {**params, "anon_phone": _ANON_PHONE},
+        )
+
+        db.execute(
```
codewithwaleed, line 107: **Incomplete GDPR erasure — Pinecone vector embeddings are not deleted.** The wipe correctly removes `kbchunk` rows from Postgres (which stores the text content and embedding vectors) and GCS recordings, but if this workspace ever used the Pinecone vector store, those embedding vectors persist in Pinecone indefinitely after this call — constituting a GDPR breach for a right-to-erasure request.

**Solution:** after the `db.commit()` and alongside the GCS cleanup, add a Pinecone namespace/vector delete:
```python
try:
    from app.services.pinecone_service import pinecone_service
    pinecone_service.delete_namespace(str(workspace_id))  # or delete_vectors_by_metadata
except Exception as exc:
    logger.error(
        "account_deletion: Pinecone cleanup failed for workspace %s: %s",
        workspace_id, exc, exc_info=True,
    )
```
If the codebase stores embeddings only in Postgres (`kbchunk.embedding`), confirm this explicitly in the PR description and add a test asserting no external vector store is called — to make the guarantee auditable.

codewithwaleed (changes requested):
Reviewed GDPR data export and right-to-erasure. The PII wipe SQL (sole-tenant user detection, audit-log anonymization, partial unique email index) is well-designed. Two issues: the DELETE endpoint passes a request body which many HTTP clients strip, and the erasure is incomplete for workspaces that used the Pinecone vector store.

app/api/v2/routers/workspace.py
Aliict: Changed DELETE-with-body to a safer approach (POST action or header-based confirmation) to avoid proxy/client issues and ensure reliable request handling.

app/services/account_deletion_service.py
Aliict: Added missing Pinecone cleanup to ensure full GDPR-compliant deletion of workspace data including external vector embeddings.

PR #107, opened 2026-06-18
feat(sdk): Web SDK public call-token endpoint with domain whitelisting

Aliict:
## Summary
- Adds `POST /api/v1/sdk/public-call-token` — the first fully unauthenticated endpoint in the app — so the Web SDK can fetch a LiveKit token without an API key. Validated via `call_flows.public_access` + a per-workspace `allowed_domains` Origin whitelist (with a `localhost:*` bypass in development), origin normalization (lowercase, no trailing slash, no `:443`), and a dedicated 20-req/min-per-IP rate limit. Returns `{livekit_token, room_name, flow_id, expires_at}` using the same Sprint-2 LiveKit token path (1h TTL). Logs origin/flow_id/IP on every request, never the token.
- `PUT /api/v1/call-flows/{id}/settings` — toggles `public_access`, gated to admin/owner.
- `POST/GET/DELETE /api/v1/workspace/allowed-domains` — HTTPS-validated, capped at 20 domains/workspace (422 past the cap).
- Migration adds `callflow.public_access` and a new `alloweddomain` table (table name follows the codebase's lowercased-classname convention rather than the ticket's literal `allowed_domains`, consistent with prior tickets on this repo).
- Added a small outermost CORS-reflection middleware scoped to only the public token route, since the static env-configured CORS policy would otherwise block a browser from any tenant-whitelisted domain that isn't also in the global `ALLOWED_ORIGINS` list. The real authorization boundary stays the `allowed_domains` DB check inside the handler — this middleware grants no security by itself.

## Test plan
- [x] Allowed origin → 200 with token
- [x] Disallowed origin → 403 `domain_not_allowed`
- [x] Flow without `public_access` → 403 `public_access_disabled`
- [x] Exceeding 20 allowed domains → 422
- [x] `localhost:*` bypass works in development without a whitelist entry
- [x] Origin normalization (case, trailing slash, default port) matches stored domains
- [x] Token value never appears in logs
- [x] IP rate-limit bucket (20/min) enforced at the middleware layer
- [x] Admin/owner-only gate on the settings endpoint (real DB-backed role check, not just a mock)
- [x] Migration applied and verified against the dev DB (`alloweddomain` table + `callflow.public_access` column present)
- [x] Full existing test suite diffed against the unmodified base branch — zero regressions (pre-existing cross-module flakiness in the full-suite run is identical with or without this branch)

### DIFF
--- a/app/core/origin.py
+++ b/app/core/origin.py
@@ -1,0 +1,24 @@
+"""Origin / domain normalization shared by allowed-domains CRUD and the
+public SDK token endpoint — both must agree on the same canonical form so
+a stored domain matches a browser's ``Origin`` header.
+
+Rules: lowercase, drop any path (Origin headers never have one anyway),
+drop the port when it's the default HTTPS port 443.
+"""
+from __future__ import annotations
+
+from urllib.parse import urlsplit
+
+
+def normalize_origin(value: str) -> str:
+    parsed = urlsplit(value.strip())
+    scheme = parsed.scheme.lower()
+    host = (parsed.hostname or "").lower()
+    port = parsed.port
+    netloc = host if not port or port == 443 else f"{host}:{port}"
+    return f"{scheme}://{netloc}"
+
+
+def is_localhost_origin(value: str) -> bool:
+    parsed = urlsplit(value.strip())
+    return (parsed.hostname or "").lower() == "localhost"
--- a/app/core/origin.py
+++ b/app/core/origin.py
@@ -19,6 +19,9 @@ def normalize_origin(value: str) -> str:
     return f"{scheme}://{netloc}"
 
 
+_LOCALHOST_HOSTS = frozenset({"localhost", "127.0.0.1", "::1"})
+
+
 def is_localhost_origin(value: str) -> bool:
     parsed = urlsplit(value.strip())
-    return (parsed.hostname or "").lower() == "localhost"
+    return (parsed.hostname or "").lower() in _LOCALHOST_HOSTS
--- a/app/services/allowed_domain_service.py
+++ b/app/services/allowed_domain_service.py
@@ -1,0 +1,69 @@
+"""Allowed-domain (Web SDK whitelist) service."""
+from __future__ import annotations
+
+import uuid
+
+from fastapi import HTTPException, status
+from sqlalchemy import func, select
+from sqlalchemy.orm import Session
+
+from app.core.origin import normalize_origin
+from app.models.allowed_domain import AllowedDomain
+from app.schemas.allowed_domain import AllowedDomainCreate, AllowedDomainOut
+
+MAX_DOMAINS_PER_WORKSPACE = 20
+
+
+class AllowedDomainService:
+    def list_domains(self, db: Session, workspace_id: uuid.UUID) -> list[AllowedDomainOut]:
+        rows = (
+            db.execute(
+                select(AllowedDomain)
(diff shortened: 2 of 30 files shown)


app/core/origin.py
```diff
@@ -0,0 +1,24 @@
...
+def normalize_origin(value: str) -> str:
+    parsed = urlsplit(value.strip())
+    scheme = parsed.scheme.lower()
+    host = (parsed.hostname or "").lower()
+    port = parsed.port
+    netloc = host if not port or port == 443 else f"{host}:{port}"
+    return f"{scheme}://{netloc}"
+
+
+def is_localhost_origin(value: str) -> bool:
```
codewithwaleed, line 25: **`is_localhost_origin` only matches `localhost`, not `127.0.0.1`.** A developer running the embedded widget from `http://127.0.0.1:3000` will be denied with a 403 in development mode, because `urlsplit("http://127.0.0.1:3000").hostname` is `'127.0.0.1'`, not `'localhost'`.

**Solution:**
```python
_LOCALHOST_HOSTS = frozenset({"localhost", "127.0.0.1", "::1"})

def is_localhost_origin(value: str) -> bool:
    parsed = urlsplit(value.strip())
    return (parsed.hostname or "").lower() in _LOCALHOST_HOSTS
```

app/services/allowed_domain_service.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+            .scalars()
+            .all()
+        )
+        return [AllowedDomainOut.model_validate(r) for r in rows]
+
+    def create_domain(
+        self, db: Session, workspace_id: uuid.UUID, body: AllowedDomainCreate
+    ) -> AllowedDomainOut:
+        count = db.execute(
```
codewithwaleed, line 32: **The domain is stored via `normalize_origin(body.domain)` (correct), but `AllowedDomainCreate._validate_https_url` validates the *raw* input — not the normalized form.** This means a user can submit `https://example.com/some-path`, the validator passes, but the stored value is `https://example.com` (path stripped). The discrepancy is silent and confusing — the returned `AllowedDomainOut.domain` will differ from what was submitted.

**Solution:** normalize in the validator and return the normalized value, so what you validate is what you store:
```python
@field_validator("domain")
@classmethod
def _validate_https_url(cls, v: str) -> str:
    parsed = urlsplit(v.strip())
    if parsed.scheme != "https" or not parsed.hostname:
        raise ValueError("domain must be a valid HTTPS URL, e.g. https://app.example.com")
    # Return the normalized form so stored value == submitted value
    from app.core.origin import normalize_origin
    return normalize_origin(v)
```
Then remove the `normalize_origin(body.domain)` call in `create_domain` since it arrives pre-normalized.

codewithwaleed (changes requested):
Reviewed the Web SDK public call-token endpoint. The per-IP rate limit, domain whitelist via `normalize_origin`, and `public_access` flag gate are all well-structured. Two issues: the localhost bypass misses `127.0.0.1`, and domain validation stores the raw submitted URL rather than the normalized form.

app/core/origin.py
Aliict: Fixed localhost validation to support localhost, 127.0.0.1, and ::1 for proper local development access.

app/services/allowed_domain_service.py
Aliict: Fixed validation/storage mismatch by ensuring domain is normalized consistently at validation time.

PR #114, opened 2026-06-25
feat: integrate HubSpot CRM via OAuth 2.0

Aliict:
## Summary
- New `workspaceintegration` table (one row per workspace+provider) stores OAuth-connected CRM accounts: `access_token`/`refresh_token` (encrypted), `token_expires_at`, `metadata`.
- HubSpot OAuth 2.0 flow: `GET /api/v1/integrations/hubspot/connect` (admin-only, redirects to HubSpot's consent screen with a signed `state` carrying the tenant id) and `GET /api/v1/integrations/hubspot/callback` (public — exempted from API-key middleware since HubSpot's redirect carries no auth headers; verifies `state`, exchanges the code, stores tokens).
- Tokens are encrypted with real **AES-256-GCM** (Python `cryptography`, not pgcrypto — `pgp_sym_encrypt` has no GCM mode). Decrypt transparently falls back to the legacy `pgp_sym_encrypt` format for tokens stored before this change, so already-connected workspaces don't need to reconnect.
- Auto-refresh: every HubSpot call goes through `get_valid_access_token`, which checks `token_expires_at` and refreshes via `refresh_token` first if needed.
- `GET /api/v1/integrations/hubspot/contact?phone=` — CRM Search API lookup by phone, Redis-cached 5 minutes per phone number, returns `{id, name, email, company, last_interaction_date}`.
- CRM context (`CRM CONTEXT: Caller name: ..., Company: ..., Last interaction: ...`) is injected into the LLM system prompt once per call (cached on `call_session.call_metadata`), fetched in `conversation_orchestrator.py` with a tight timeout — fails open, never blocks the call.
- Post-call write-back: when a call transitions to `status='completed'` (hooked into `CallSessionService.update_call_session_status`, same pattern as the existing inbound-CRM-sync/smart-callback hooks), a fire-and-forget background task generates a 2-sentence Gemini summary of the transcript and creates a HubSpot Call engagement (duration, outcome, summary) associated with the contact.
- `DELETE /api/v1/integrations/hubspot` revokes the refresh token at HubSpot and deletes the local row. `GET /api/v1/integrations` now reports `{name: "hubspot", connected, connected_at}` alongside Make/n8n.
- All HubSpot HTTP calls go through a shared backoff helper (110 req/10s rate limit; exponential backoff on 429, honoring `Retry-After`).

## Test plan
- [x] 99 pytest tests pass: OAuth state sign/verify, token refresh (fresh/expired/no-refresh-token/refresh-failure), contact shape + Redis cache hit/miss/fail-open, CRM context injection + caching, post-call write-back, disconnect, 429 backoff, `GET /integrations` HubSpot entry, OpenAPI contract
- [x] Migration verified both ways (upgrade/downgrade) against a throwaway local Postgres container, then applied to the dev database
- [x] AES-256-GCM round-trip and legacy-token backward compatibility verified directly against the dev DB's already-connected row
- [x] Manually connected a real HubSpot account end-to-end via `/connect` → consent → `/callback`

### DIFF
--- a/app/core/db_encryption.py
+++ b/app/core/db_encryption.py
@@ -230,7 +230,18 @@ _HUBSPOT_AESGCM_PREFIX = "gcm1:"
 
 
 def _hubspot_aes_key() -> bytes:
-    """Derive a 32-byte AES-256 key from HUBSPOT_TOKEN_ENCRYPTION_KEY (any length/format)."""
+    """Derive a 32-byte AES-256 key from HUBSPOT_TOKEN_ENCRYPTION_KEY.
+
+    SECURITY NOTE:
+      For secure AES-256-GCM encryption, HUBSPOT_TOKEN_ENCRYPTION_KEY must be configured
+      as a randomly generated 32-byte secret (typically represented as a 64-character
+      hex string, e.g. generated via `secrets.token_hex(32)`).
+
+    BACKWARD COMPATIBILITY:
+      We derive the key using hashlib.sha256 to ensure that any key length/format can
+      be safely resolved, and to preserve 100% backward compatibility for already
+      connected HubSpot workspaces encrypted under older/existing settings keys.
+    """
     key = settings.HUBSPOT_TOKEN_ENCRYPTION_KEY
     if not key:
         raise ValueError(
--- a/app/core/db_encryption.py
+++ b/app/core/db_encryption.py
@@ -13,14 +13,22 @@ used by the Twilio credentials.
 
 If the key is missing or empty the encrypt call raises ``ValueError`` so that
 the application fails loudly rather than silently storing plaintext.
+
+HubSpot tokens (below) are the one exception: pgcrypto's ``pgp_sym_encrypt``
+is OpenPGP symmetric encryption and has no GCM mode, so it cannot satisfy a
+literal AES-256-GCM requirement. Those two helpers perform AES-256-GCM in
+Python via ``cryptography`` instead of pgcrypto SQL.
 """
 
 from __future__ import annotations
 
 import base64
+import hashlib
 import logging
+import os
 from contextlib import contextmanager
 
+from cryptography.hazmat.primitives.ciphers.aead import AESGCM
 from sqlalchemy import text
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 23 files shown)


app/services/hubspot_service.py
```diff
@@ -0,0 +1,599 @@
...
+    db.delete(row)
+    db.commit()
+    return True
+
+
+# ─── Contact lookup (CRM Search API, Redis-cached) ────────────────────────────
+
+
+def _contact_cache_key(tenant_id: uuid.UUID, phone: str) -> str:
+    return f"hubspot:contact:{tenant_id}:{phone}"
```
codewithwaleed: **Phone number stored in plaintext as part of the Redis cache key — PII exposure.** The key `hubspot:contact:{tenant_id}:{phone}` writes the caller's phone number directly into Redis. Anyone with Redis access (ops, monitoring tools like RedisInsight/DataDog, `KEYS *` in CLI) can enumerate every phone number that ever triggered a CRM lookup.

**Solution:** hash the phone before using it as the key suffix:
```python
import hashlib

def _contact_cache_key(tenant_id: uuid.UUID, phone: str) -> str:
    phone_hash = hashlib.sha256(phone.encode()).hexdigest()
    return f"hubspot:contact:{tenant_id}:{phone_hash}"
```
Same input → same key, so cache hits still work, but the phone number is never stored in plaintext in Redis.

app/services/hubspot_service.py
```diff
@@ -0,0 +1,599 @@
...
+        "POST",
+        SEARCH_CONTACTS_URL,
+        headers={"Authorization": f"Bearer {access_token}"},
+        json={
+            "filterGroups": [
+                {
+                    "filters": [
+                        {
+                            "propertyName": "phone",
+                            "operator": "CONTAINS_TOKEN",
```
codewithwaleed: **`CONTAINS_TOKEN` does partial/prefix matching — will silently return the wrong contact when phone numbers overlap.** HubSpot's `CONTAINS_TOKEN` operator matches any property that *contains* the search string as a token, so searching for `+14155551234` can match a contact whose stored number is `+141555512345`. In a voice context this means the wrong caller's history and name are injected into the agent's system prompt.

**Solution:** use `EQ` for exact matching:
```python
{
    "propertyName": "phone",
    "operator": "EQ",
    "value": phone,
}
```
If you need to handle both `(415) 555-1234` and `+14155551234` formats, normalize to E.164 before the call (Python `phonenumbers` lib). You can add a second `filterGroup` with `mobilephone` to cover contacts whose number is stored under the mobile field.

app/services/hubspot_service.py
```diff
@@ -0,0 +1,599 @@
...
+        )
+        context_block = ""
+
+    try:
+        updated_metadata = dict(call_session.call_metadata or {})
+        updated_metadata["hubspot_crm_context"] = context_block
+        call_session.call_metadata = updated_metadata
+        flag_modified(call_session, "call_metadata")
+        db.add(call_session)
+        db.commit()
```
codewithwaleed: **`db.commit()` inside `get_crm_context_block_for_call` fires mid-turn during LLM prompt construction.** This helper is awaited from `conversation_orchestrator.py` inside the per-turn prompt-building block. Calling `db.commit()` here flushes *all* dirty ORM state on the session — including any transcript rows or call-session fields staged by other parts of the same turn — before the LLM response has even returned. If the LLM call fails afterwards, those changes are already committed and unrollbackable.

**Solution:** use `db.flush()` instead to push the metadata update to the in-progress transaction without committing, and let the outer request handler own the transaction boundary:
```python
try:
    updated_metadata = dict(call_session.call_metadata or {})
    updated_metadata["hubspot_crm_context"] = context_block
    call_session.call_metadata = updated_metadata
    flag_modified(call_session, "call_metadata")
    db.add(call_session)
    db.flush()  # ← not db.commit()
except Exception:
    db.rollback()
    logger.warning("Failed to cache HubSpot CRM context (non-critical)", exc_info=True)
```

app/core/db_encryption.py
```diff
@@ -13,14 +13,22 @@
...
+Python via ``cryptography`` instead of pgcrypto SQL.
 """
 
 from __future__ import annotations
 
 import base64
+import hashlib
 import logging
+import os
 from contextlib import contextmanager
```
codewithwaleed, line 29: **SHA-256 used as a KDF — safe only if `HUBSPOT_TOKEN_ENCRYPTION_KEY` is already a high-entropy random value.** `hashlib.sha256(key.encode()).digest()` applies a single hash pass with no salt or iterations. If an operator sets the env var to a short passphrase, the derived 256-bit key has the same low entropy.

**Recommendation:** document in `.env.example` that this must be a randomly-generated 32-byte secret, not a passphrase:
```bash
# Generate: python -c "import secrets; print(secrets.token_hex(32))"
HUBSPOT_TOKEN_ENCRYPTION_KEY=<64 random hex chars>
```
Or switch to HKDF to be safe regardless of what operators supply:
```python
from cryptography.hazmat.primitives.kdf.hkdf import HKDF
from cryptography.hazmat.primitives import hashes

def _hubspot_aes_key() -> bytes:
    raw = settings.HUBSPOT_TOKEN_ENCRYPTION_KEY
    if not raw:
        raise ValueError("HUBSPOT_TOKEN_ENCRYPTION_KEY is not configured")
    return HKDF(
        algorithm=hashes.SHA256(), length=32, salt=None, info=b"hubspot-token-enc"
    ).derive(raw.encode("utf-8"))
```

codewithwaleed (changes requested):
Reviewed the HubSpot OAuth 2.0 integration. The AES-256-GCM token encryption with legacy pgcrypto fallback is well designed, the signed JWT state for the OAuth round-trip is solid, and the fail-open pattern throughout the call path is correct. Three issues need fixing before merge: phone PII in Redis keys, an inexact phone-search operator that will cause wrong-contact matches, and a mid-turn DB commit during LLM prompt construction.

app/services/hubspot_service.py
Aliict: Updated _contact_cache_key() to construct Redis keys using a deterministic SHA-256 hash of the phone number instead of storing phone numbers in plaintext.

app/services/hubspot_service.py
Aliict: Created normalize_to_e164() helper to normalize incoming phone number formats.
Created _get_phone_search_values() to generate unique exact-match variations (E.164, raw input, digits only, 10-digit national format for US numbers).
Updated search_contact_by_phone() to search both phone and mobilephone properties against these variations using the exact-match EQ operator instead of CONTAINS_TOKEN, preventing partial matches.
Updated get_contact_for_phone() to normalize phone numbers before caching or searching.

app/services/hubspot_service.py
Aliict: Updated context caching in get_crm_context_block_for_call(). Replaced db.commit() with db.flush() inside a nested transaction savepoint (db.begin_nested()). This ensures that context block generation during LLM prompt building fails open without committing database transactions or corrupting active transaction states.

app/core/db_encryption.py
Aliict: Added clear documentation indicating that HUBSPOT_TOKEN_ENCRYPTION_KEY must be a randomly generated 32-byte secret (represented as a 64-char hex string) to guarantee security. Retained the current SHA-256 key derivation process to maintain 100% backward compatibility for already connected/encrypted workspaces.

PR #124, opened 2026-07-02
feat: HubSpot field mapping, settings toggles, and write-back hardening

Aliict:
## Summary
- Adds tenant-configurable HubSpot field -> agent prompt-variable mappings (`POST /api/v1/integrations/hubspot/field-mapping`), resolved from the CRM Search API response and substituted into the system prompt (`{prompt_variable}` tokens) once per call, cached on `call_session.call_metadata`.
- Adds `GET /api/v1/integrations/hubspot` (connection status + toggles + mappings) and `PUT /api/v1/integrations/hubspot/settings` to toggle `contact_lookup_enabled` / `write_back_enabled` per workspace, stored in `WorkspaceIntegration.extra_metadata`.
- Post-call write-back now respects `write_back_enabled`, caches the 2-sentence Gemini transcript summary per call (avoids duplicate Gemini calls on retry), and records/clears `last_write_back_error` on `WorkspaceIntegration.extra_metadata` instead of silently swallowing failures — no auto-retry.
- Contact lookup now requests any tenant-mapped HubSpot properties alongside the existing default set, without changing the shape of the existing `HubSpotContactOut` response.

## Test plan
- [x] `pytest tests/services/test_hubspot_service.py tests/api/test_hubspot_integration.py -q` — 75 passed
- [x] Full suite (`pytest tests/ -q`) run for regressions — 139 pre-existing failures confirmed unrelated (reproduced identically on `dev` before this branch's changes; missing external API keys/creds in this environment, not touched by this change)
- [x] Manual: connect a HubSpot sandbox account, set field mappings + toggles via the new endpoints, place a test call, and confirm prompt substitution + write-back behavior end-to-end

### DIFF
--- a/app/schemas/hubspot_integration.py
+++ b/app/schemas/hubspot_integration.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import Optional
+from datetime import datetime
+from typing import List, Optional
 
 from pydantic import BaseModel
 
@@ -20,3 +21,29 @@ class HubSpotContactOut(BaseModel):
 class HubSpotDisconnectResponse(BaseModel):
     disconnected: bool
     provider: str = "hubspot"
+
+
+class HubSpotFieldMapping(BaseModel):
+    hubspot_field: str
+    prompt_variable: str
+
+
+class HubSpotFieldMappingRequest(BaseModel):
+    mappings: List[HubSpotFieldMapping]
+
+
+class HubSpotFieldMappingResponse(BaseModel):
+    field_mappings: List[HubSpotFieldMapping]
+
+
+class HubSpotSettingsUpdateRequest(BaseModel):
+    contact_lookup_enabled: bool
+    write_back_enabled: bool
+
+
+class HubSpotIntegrationStatusOut(BaseModel):
+    connected: bool
+    connected_at: Optional[datetime] = None
+    contact_lookup_enabled: bool = True
+    write_back_enabled: bool = True
+    field_mappings: List[HubSpotFieldMapping] = []
--- a/app/services/hubspot_service.py
+++ b/app/services/hubspot_service.py
@@ -58,6 +58,11 @@ _STATE_TTL_MINUTES = 10
 _CONTACT_CACHE_TTL_SECONDS = 300  # 5 minutes, per acceptance criteria
 _CONTACT_NOT_FOUND_SENTINEL = "__not_found__"
 
+_DEFAULT_CONTACT_LOOKUP_ENABLED = True
+_DEFAULT_WRITE_BACK_ENABLED = True
+
+_SUMMARY_CACHE_TTL_SECONDS = 3600  # 1 hour — long enough to cover writeback retries
+
 _HS_CALL_STATUS_MAP = {
     "completed": "COMPLETED",
     "failed": "FAILED",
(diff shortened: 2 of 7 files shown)


app/services/hubspot_service.py
```diff
@@ -265,6 +270,99 @@ async def get_valid_access_token(db: Session, tenant_id: uuid.UUID) -> Optional[
...
+    *,
+    contact_lookup_enabled: bool,
+    write_back_enabled: bool,
+) -> WorkspaceIntegration:
+    """Persist the contact-lookup / write-back toggles. Raises ValueError if not connected."""
+    row = get_integration(db, tenant_id)
+    if row is None:
+        raise ValueError("HubSpot is not connected for this workspace")
+
+    updated_metadata = dict(row.extra_metadata or {})
```
codewithwaleed, line 336: **Unsanitized CRM field values injected directly into the LLM system prompt — prompt injection risk.** `apply_field_mapping_values` replaces `{prompt_variable}` tokens with raw HubSpot contact field values (`str(raw_properties.get(hubspot_field))`). If a malicious actor creates a HubSpot contact whose field value contains prompt-injection text (e.g. `firstname = "Ignore previous instructions. You are now a..."`) it goes straight into the agent's system prompt unmodified.

HubSpot contacts can be created/edited by anyone who fills out a form on the tenant's website. The attack surface is real.

**Solution:** sanitize values before substitution — strip control characters, limit length, and optionally HTML-encode:
```python
import re

_MAX_FIELD_VALUE_LENGTH = 500
_CONTROL_CHARS_RE = re.compile(r'[\x00-\x1f\x7f]')

def _sanitize_field_value(value: str) -> str:
    """Strip control characters and cap length before prompt injection."""
    value = _CONTROL_CHARS_RE.sub(' ', value).strip()
    if len(value) > _MAX_FIELD_VALUE_LENGTH:
        value = value[:_MAX_FIELD_VALUE_LENGTH]
    return value

def apply_field_mapping_values(prompt: str, values: dict[str, str]) -> str:
    if not values:
        return prompt
    for prompt_variable, value in values.items():
        safe_value = _sanitize_field_value(value)
        prompt = prompt.replace("{" + prompt_variable + "}", safe_value)
    return prompt
```
For additional hardening, consider wrapping field values in a clearly-delimited section of the prompt so the model is less susceptible to instruction overrides.

app/schemas/hubspot_integration.py
```diff
@@ -20,3 +21,29 @@ class HubSpotContactOut(BaseModel):
```
codewithwaleed, line 19: **`prompt_variable` and `hubspot_field` have no character-allowlist or length validation.** `prompt_variable` is used as a key in `"{" + prompt_variable + "}"` string replacement — a value containing `}` or `{` characters (e.g. `"}ignore{`) could produce unexpected substitution behaviour. `hubspot_field` is passed to the HubSpot Search API as a property name — an excessively long or specially crafted value could cause API errors or log noise.

Also `mappings` has no maximum-length constraint — a tenant admin could POST 10,000 mappings, which would be stored and then queried against HubSpot on every call leg.

**Fix:** add validators to `HubSpotFieldMapping` and the request:
```python
from pydantic import BaseModel, field_validator, model_validator
import re

_IDENTIFIER_RE = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]{0,127}$')
_HUBSPOT_FIELD_RE = re.compile(r'^[a-z_][a-z0-9_]{0,127}$')

class HubSpotFieldMapping(BaseModel):
    hubspot_field: str
    prompt_variable: str

    @field_validator('hubspot_field')
    @classmethod
    def validate_hubspot_field(cls, v: str) -> str:
        if not _HUBSPOT_FIELD_RE.match(v):
            raise ValueError('hubspot_field must be a valid HubSpot property name (lowercase, underscores, max 128 chars)')
        return v

    @field_validator('prompt_variable')
    @classmethod
    def validate_prompt_variable(cls, v: str) -> str:
        if not _IDENTIFIER_RE.match(v):
            raise ValueError('prompt_variable must be a valid identifier (letters, digits, underscores, max 128 chars)')
        return v

class HubSpotFieldMappingRequest(BaseModel):
    mappings: List[HubSpotFieldMapping]

    @model_validator(mode='after')
    def validate_mapping_count(self) -> 'HubSpotFieldMappingRequest':
        if len(self.mappings) > 50:
            raise ValueError('A maximum of 50 field mappings are allowed')
        return self
```

app/services/hubspot_service.py
```diff
@@ -265,6 +270,99 @@ async def get_valid_access_token(db: Session, tenant_id: uuid.UUID) -> Optional[
...
+    updated_metadata["field_mappings"] = mappings
+    row.extra_metadata = updated_metadata
+    flag_modified(row, "extra_metadata")
+    db.add(row)
+    db.commit()
```
codewithwaleed, line 293: **`str(exc)` of a HubSpot API error is persisted to the DB and returned to the client via `GET /integrations/hubspot`.** HubSpot API errors can contain URLs, auth headers, or internal API response bodies depending on how `httpx` formats exceptions. While the tenant dashboard is the consumer (not a public surface), storing raw exception strings as user-visible state is fragile.

**Fix:** cap and sanitize the error message before storing:
```python
_MAX_ERROR_LEN = 500

def _safe_error_msg(exc: Exception) -> str:
    """Extract a brief, safe error summary for tenant-visible storage."""
    msg = str(exc)
    # Redact anything that looks like a bearer token
    msg = re.sub(r'Bearer [A-Za-z0-9\-._~+/]+=*', 'Bearer [redacted]', msg)
    return msg[:_MAX_ERROR_LEN]

# In _run_post_call_writeback_async:
set_last_write_back_error(db, tenant_id, _safe_error_msg(exc))
```

codewithwaleed (changes requested):
Good additions overall — the settings toggles, field-mapping persistence, and cached transcript summary are all solid. The `begin_nested()` + `flush()` pattern for mid-call metadata writes is correct and addresses the commit-boundary issue from PR #114. Two issues need fixing before merge: unsanitized HubSpot field values injected into the LLM system prompt (prompt injection), and no length/character validation on `prompt_variable` values stored by admins.

PR #125, opened 2026-07-02
feat: A/B prompt testing for call flows

Aliict:
## Summary

Adds an A/B prompt testing system for call flows: traffic can be split between two `PromptVersion`s on a flow, results are measured per variant, and a winner can be promoted to the flow's active prompt.

### Data model
- `CallFlow`: `ab_test_enabled` (bool), `ab_prompt_a_id` / `ab_prompt_b_id` (FK → `promptversion.id`), `ab_split_ratio` (Numeric(3,2), constrained to `0 < ratio < 1`, default `0.50`).
- `CallSession`: `ab_variant` (`'a'` | `'b'` | `NULL`) — the variant locked in for that call.
- Migration `20260702_ab_prompt_testing` (applied to the dev DB already).

### How prompt resolution works now
1. **Assignment (dispatch time, outbound calls only)** — in `voice_call_service.py`, right after `call_flow_id` is stamped onto the new `CallSession` and *before* any LiveKit/Twilio/LLM call is made, `ab_testing_service.assign_and_lock_variant()`:
   - Checks the flow's `ab_test_enabled` + both prompt IDs are set.
   - Rolls `random.random() < ab_split_ratio` → `'a'` or `'b'`.
   - Resolves the winning `PromptVersion.prompt_text` and writes both the variant and the resolved text onto the `CallSession` immediately (`ab_variant` column + `call_metadata["ab_prompt_text"]`), committed right away — so the variant is known even if the call fails a moment later.
   - Inbound calls don't currently associate a `CallFlow` with the session at all in this codebase, so A/B assignment only applies to the outbound dispatch path today.
2. **Resolution (live call, locked for the call's duration)** — both `bidirectional_stream.py` and `conversation_orchestrator.py` (the two places that build the LLM system prompt) now read `call_metadata["ab_prompt_text"]` and substitute it in place of `agent.system_prompt` / `agent.model.system_prompt`, using the exact same override seam that already existed for `batch_prompt_override` (batch override still wins if both happen to be set on the same call). Because the value is read once from `call_metadata` and never re-queried, the same variant's prompt is used for the entire call — no mid-call switching.

### API (`/api/v2/flows`)
- `PUT /{flow_id}/ab-test` — enable/configure A/B testing; validates both prompt IDs belong to the flow and `split_ratio` is in `[0.1, 0.9]`.
- `GET /{flow_id}/ab-results` — per-variant `calls`, `completed`, `failed`, `avg_duration`, `transfer_rate`, `success_rate`, plus a chi-squared significance test (`scipy.stats.chi2_contingency`) on completed-vs-total; guardrail: either variant under 30 calls → always `inconclusive`, otherwise significant only at `p < 0.05`, recommending the variant with the higher completion rate.
- `PUT /{flow_id}/ab-test/winner` — sets the flow's `current_prompt_id` to the winning variant's prompt and disables the test.

### Call history
- `CallHistoryItem.ab_variant` added and populated from `CallSession.ab_variant` in `call_history_service.get_list`.

## Test plan
- [x] `tests/api/test_ab_prompt_testing.py` (14 tests): random-split tolerance over 1000 samples, disabled/incomplete-config no-ops, `ab-test` config validation (400s for bad split ratio / prompt not on flow), `ab-results` metric math, 30-call significance guardrail, significant vs. inconclusive chi-squared outcomes, winner promotion (+ 400 on incomplete variant config).
- [x] Full suite run: 1344 passed; the 32 pre-existing failures were verified identical on the base branch (unrelated: contact recovery, Google STT, rate limiting, TTS/barge-in tests, a stale mock target).
- [x] Migration applied to dev DB and columns verified via `information_schema`.

### DIFF
--- a/app/schemas/ab_testing.py
+++ b/app/schemas/ab_testing.py
@@ -1,0 +1,48 @@
+from __future__ import annotations
+
+import uuid
+from typing import Literal, Optional
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class AbTestUpdate(BaseModel):
+    """Request body for ``PUT /api/v2/flows/{flow_id}/ab-test``."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    enabled: bool
+    prompt_a_id: uuid.UUID
+    prompt_b_id: uuid.UUID
+    split_ratio: float = Field(..., ge=0.1, le=0.9)
+
+
+class AbTestResponse(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+
+    ab_test_enabled: bool
+    ab_prompt_a_id: Optional[uuid.UUID]
+    ab_prompt_b_id: Optional[uuid.UUID]
+    ab_split_ratio: float
+
+
+class VariantMetrics(BaseModel):
+    calls: int
+    completed: int
+    failed: int
+    avg_duration: Optional[float]
+    transfer_rate: float
+    success_rate: float
+
+
+class AbResultsResponse(BaseModel):
+    variant_a: VariantMetrics
+    variant_b: VariantMetrics
+    statistical_significance: bool
+    recommended_variant: Literal["a", "b", "inconclusive"]
+
+
+class AbTestWinnerUpdate(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    variant: Literal["a", "b"]
--- a/app/services/ab_testing_service.py
+++ b/app/services/ab_testing_service.py
@@ -1,0 +1,79 @@
+"""A/B prompt testing — variant assignment at dispatch time and results aggregation.
+
+Random assignment happens once, at call-session creation, and is persisted onto
+``CallSession.ab_variant`` (plus the resolved prompt text in ``call_metadata``)
+before any LLM request is made — so the variant is always known even if the
+call fails mid-way, and never changes for the lifetime of the call.
+"""
+from __future__ import annotations
+
+import random
+from typing import Optional
+
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from app.core.logger import logger
+from app.models.call_flow import CallFlow
+from app.models.call_session import CallSession
+from app.models.prompt_version import PromptVersion
+
+
+class AbTestingService:
+    def pick_variant(self, call_flow: CallFlow) -> Optional[str]:
+        """Return 'a' or 'b' if the flow is eligible for A/B testing, else None."""
+        if not call_flow.ab_test_enabled:
+            return None
+        if not call_flow.ab_prompt_a_id or not call_flow.ab_prompt_b_id:
+            return None
+        split_ratio = float(call_flow.ab_split_ratio)
+        return "a" if random.random() < split_ratio else "b"
+
+    def get_variant_prompt_text(
+        self, db: Session, call_flow: CallFlow, variant: str
+    ) -> Optional[str]:
+        prompt_id = (
+            call_flow.ab_prompt_a_id if variant == "a" else call_flow.ab_prompt_b_id
+        )
+        if prompt_id is None:
+            return None
+        version = db.execute(
+            select(PromptVersion).where(PromptVersion.id == prompt_id)
+        ).scalar_one_or_none()
+        return version.prompt_text if version else None
+
+    def assign_and_lock_variant(
+        self, db: Session, call_session: CallSession, call_flow: CallFlow
+    ) -> None:
+        """Assign a variant (if the flow has A/B testing enabled) and persist it
+        onto the call session immediately, before any LLM request."""
+        variant = self.pick_variant(call_flow)
+        if variant is None:
+            return
+
+        prompt_text = self.get_variant_prompt_text(db, call_flow, variant)
+        if prompt_text is None:
+            logger.warning(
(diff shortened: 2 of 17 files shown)


app/services/call_flow_service.py
```diff
@@ -367,5 +377,160 @@ def get_prompt_versions(
...
+                "ab_prompt_a_id": body.prompt_a_id,
+                "ab_prompt_b_id": body.prompt_b_id,
+                "ab_split_ratio": body.split_ratio,
+            },
+        )
+        db.commit()
+        db.refresh(flow)
+        return AbTestResponse(
+            ab_test_enabled=flow.ab_test_enabled,
+            ab_prompt_a_id=flow.ab_prompt_a_id,
```
codewithwaleed, line 419: **`_variant_metrics` loads all call sessions for a flow into memory and aggregates in Python — will OOM on any flow with significant call volume.** `db.execute(select(CallSession).where(...)).scalars().all()` fetches every `CallSession` row for the flow + variant into the web-server process. A flow with 50,000 calls has 50,000 SQLAlchemy objects materialized in RAM; `GET /ab-results` becomes a memory spike every time someone opens the dashboard.

**Solution:** push all aggregation into SQL using `func` expressions:
```python
from sqlalchemy import case, func

def _variant_metrics(
    self,
    db: Session,
    flow_id: uuid.UUID,
    tenant_id: uuid.UUID,
    variant: str,
) -> VariantMetrics:
    row = db.execute(
        select(
            func.count().label('calls'),
            func.sum(
                case((CallSession.status == 'completed', 1), else_=0)
            ).label('completed'),
            func.sum(
                case((CallSession.status == 'failed', 1), else_=0)
            ).label('failed'),
            func.avg(CallSession.duration).label('avg_duration'),
            func.sum(
                case((CallSession.transferred == True, 1), else_=0)  # noqa: E712
            ).label('transferred'),
            func.sum(
                case((CallSession.success_evaluation == 'success', 1), else_=0)
            ).label('successes'),
        ).where(
            CallSession.call_flow_id == flow_id,
            CallSession.tenant_id == tenant_id,
            CallSession.ab_variant == variant,
        )
    ).one()

    calls = row.calls or 0
    return VariantMetrics(
        calls=calls,
        completed=row.completed or 0,
        failed=row.failed or 0,
        avg_duration=float(row.avg_duration) if row.avg_duration else None,
        transfer_rate=(row.transferred / calls) if calls else 0.0,
        success_rate=(row.successes / calls) if calls else 0.0,
    )
```
This is a single aggregation query instead of N object fetches, and handles any volume without memory pressure.

app/schemas/ab_testing.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+
+import uuid
+from typing import Literal, Optional
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class AbTestUpdate(BaseModel):
```
codewithwaleed, line 9: **No validation that `prompt_a_id != prompt_b_id`** — if both variants point to the same prompt version, the A/B test is meaningless (every call gets the identical prompt) but runs silently without any error, wasting traffic and producing nonsense results.

**Fix:** add a `model_validator`:
```python
from pydantic import BaseModel, ConfigDict, Field, model_validator

class AbTestUpdate(BaseModel):
    model_config = ConfigDict(extra="forbid")

    enabled: bool
    prompt_a_id: uuid.UUID
    prompt_b_id: uuid.UUID
    split_ratio: float = Field(..., ge=0.1, le=0.9)

    @model_validator(mode='after')
    def prompts_must_differ(self) -> 'AbTestUpdate':
        if self.prompt_a_id == self.prompt_b_id:
            raise ValueError('prompt_a_id and prompt_b_id must be different prompt versions')
        return self
```

app/services/ab_testing_service.py
```diff
@@ -0,0 +1,79 @@
...
+                variant,
+            )
+            return
+
+        call_session.ab_variant = variant
+        call_session.call_metadata = {
+            **(call_session.call_metadata or {}),
+            "ab_prompt_text": prompt_text,
+        }
+        db.commit()
```
codewithwaleed, line 69: **`db.commit()` inside `assign_and_lock_variant` — the service owns the transaction boundary when the caller should.** This is the same pattern flagged in PR #114 review: a service method committing mid-call-setup means that if any code in `voice_call_service.py` after line 730 fails, the A/B variant is already durably committed but the call may be in an inconsistent state (e.g. partial session setup).

**Fix:** use `db.flush()` instead of `db.commit()` so the write is staged but the caller retains the ability to rollback the entire setup atomically:
```python
call_session.ab_variant = variant
call_session.call_metadata = {
    **(call_session.call_metadata or {}),
    "ab_prompt_text": prompt_text,
}
db.flush()  # ← stage the write; let the caller's transaction commit
db.refresh(call_session)
```
The caller in `voice_call_service.py` does its own `db.commit()` shortly after (line 718), which will commit the flush along with all other session changes atomically.

codewithwaleed (changes requested):
A/B testing feature is well-structured — the variant-lock-before-LLM pattern, chi-squared significance test, and `promote_ab_winner` endpoint are all solid design choices. Two issues need fixing before merge: a scalability bug in metrics aggregation that will OOM on large flows, and a missing validation that lets both variants point to the same prompt.

PR #126, opened 2026-07-02
feat: cross-session caller memory for voice agent call flows

Aliict:
## Summary
- Adds cross-session caller memory: on each call, injects summaries of the caller's previous completed calls (matched by tenant, call flow, and `from_number`) into the LLM system prompt, right after the KB/CRM context blocks and before conversation history.
- New per-flow settings `caller_memory_enabled` / `caller_memory_window` (1–10, default 3) on `CallFlow`, plus `CallSession.transcript_summary` populated from the existing end-of-call HIPAA-redacted analysis.
- Lookup is fail-open: wrapped in a 100ms `asyncio.wait_for` timeout and cached on `call_session.call_metadata["caller_memory_context"]` so later turns in the same call are a cheap dict read, not a repeat query.
- `PUT /api/v2/flows/{flow_id}/caller-memory-settings` (admin-rank, audited) to configure the feature per flow. Deliberately **not** `/flows/{flow_id}/settings` — that path is already owned by the HIPAA compliance toggle in `hipaa.py`, and reusing it would have silently shadowed that endpoint.
- `GET /api/v2/calls/{call_id}/memory-context` debug endpoint to inspect the cached context for a given call, tenant-isolated.
- Migration `20260703_caller_memory` adds the new columns, a `caller_memory_window` check constraint (1–10), and a composite lookup index `ix_callsession_memory_lookup(tenant_id, call_flow_id, from_number, start_time)`. Already applied to the shared dev database.
- `docs/api/openapi.yaml` regenerated to include the two new endpoints.

## Test plan
- [x] `pytest tests/services/test_caller_memory_service.py` — cache hit/miss, fail-open on timeout/exception, real-SQL filter/order/limit query (14 tests)
- [x] `pytest tests/services/test_voice_analysis_service.py` — `transcript_summary` persisted from end-of-call analysis (2 tests)
- [x] `pytest tests/api/v2/test_caller_memory_settings.py` — RBAC, window validation, tenant isolation, audit event, route-collision sanity check (11 tests)
- [x] `pytest tests/api/v2/test_caller_memory_debug.py` — success, empty cache, cross-tenant 404, unknown-call 404 (5 tests)
- [x] Ran alongside `test_call_flows.py`, `test_ab_prompt_testing.py`, `test_hipaa.py`, `test_callback_scheduler.py` together — no regressions (196 tests passing)
- [x] `pytest tests/test_openapi_contract.py` passes against the regenerated spec
- [x] Migration applied and verified against the target Postgres database (columns, index, and check constraint confirmed present)

### DIFF
--- a/app/services/caller_memory_service.py
+++ b/app/services/caller_memory_service.py
@@ -1,0 +1,123 @@
+"""
+Cross-session caller memory: injects summaries of a caller's previous completed
+calls into the LLM system prompt so agents recall prior interactions.
+
+Retrieval is fetched once per call and cached on call_session.call_metadata
+(mirrors get_crm_context_block_for_call in hubspot_service.py) so later turns
+in the same call — the prompt is rebuilt every turn — reuse the cached string
+instead of re-querying the database. Fails open on timeout/error: a slow or
+broken lookup never blocks the call, it just proceeds without caller memory.
+"""
+from __future__ import annotations
+
+import asyncio
+from typing import List, Optional
+
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+from sqlalchemy.orm.attributes import flag_modified
+
+from app.core.config import settings
+from app.core.logger import logger
+from app.models.call_flow import CallFlow
+from app.models.call_session import CallSession
+
+_CACHE_KEY = "caller_memory_context"
+_DEFAULT_FETCH_TIMEOUT_SEC = 0.1
+
+
+def _fetch_recent_summaries(
+    db: Session, call_session: CallSession, window: int
+) -> List[CallSession]:
+    """Blocking DB query — run inside a threadpool executor with a timeout."""
+    stmt = (
+        select(CallSession)
+        .where(
+            CallSession.tenant_id == call_session.tenant_id,
+            CallSession.call_flow_id == call_session.call_flow_id,
+            CallSession.from_number == call_session.from_number,
+            CallSession.status == "completed",
+            CallSession.id != call_session.id,
+            CallSession.transcript_summary.isnot(None),
+            CallSession.transcript_summary != "",
+        )
+        .order_by(CallSession.start_time.desc())
+        .limit(window)
+    )
+    return list(db.execute(stmt).scalars().all())
+
+
+def _format_caller_memory_block(sessions: List[CallSession]) -> str:
+    if not sessions:
+        return ""
+
+    lines = [f"CALLER HISTORY (last {len(sessions)} interactions):"]
+    for session in sessions:
+        date_str = (
+            session.start_time.strftime("%Y-%m-%d") if session.start_time else "unknown date"
+        )
+        summary = (session.transcript_summary or "").strip()
+        lines.append(f"- Call on {date_str}: {summary}")
+    lines.append("End of caller history.")
+    return "\n".join(lines)
+
+
+async def get_caller_memory_context_block_for_call(
+    db: Session,
+    call_session: Optional[CallSession],
+    call_flow: Optional[CallFlow],
+) -> str:
+    """
+    Fetch the caller-memory context block for a call's system prompt, once per call.
+
+    Returns "" (no block injected) when: caller memory is disabled on the flow,
+    the call has no from_number, the lookup times out, or the lookup fails.
+    """
+    if call_session is None or call_flow is None or not call_flow.caller_memory_enabled:
+        return ""
+    if not call_session.from_number:
+        return ""
+
+    metadata = call_session.call_metadata or {}
+    if _CACHE_KEY in metadata:
+        return metadata.get(_CACHE_KEY) or ""
+
+    window = call_flow.caller_memory_window
+    context_block = ""
+    try:
+        loop = asyncio.get_running_loop()
+        sessions = await asyncio.wait_for(
+            loop.run_in_executor(None, _fetch_recent_summaries, db, call_session, window),
+            timeout=float(
+                getattr(settings, "VOICE_CALLER_MEMORY_FETCH_TIMEOUT_SEC", None)
+                or _DEFAULT_FETCH_TIMEOUT_SEC
+            ),
+        )
+        context_block = _format_caller_memory_block(sessions)
+    except asyncio.TimeoutError:
+        logger.warning(
+            "caller_memory lookup timed out after %sms; proceeding without caller memory",
+            int(_DEFAULT_FETCH_TIMEOUT_SEC * 1000),
+        )
+    except Exception:
+        logger.warning(
+            "caller_memory lookup failed; proceeding without caller memory", exc_info=True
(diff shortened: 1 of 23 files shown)


app/services/caller_memory_service.py
```diff
@@ -0,0 +1,123 @@
...
+    metadata = call_session.call_metadata or {}
+    if _CACHE_KEY in metadata:
+        return metadata.get(_CACHE_KEY) or ""
+
+    window = call_flow.caller_memory_window
+    context_block = ""
+    try:
+        loop = asyncio.get_running_loop()
+        sessions = await asyncio.wait_for(
+            loop.run_in_executor(None, _fetch_recent_summaries, db, call_session, window),
```
codewithwaleed: **`loop.run_in_executor(None, _fetch_recent_summaries, db, call_session, window)` passes a SQLAlchemy `Session` and an ORM-bound object to a thread-pool worker while the async event loop continues using the same session — thread-safety violation.** SQLAlchemy `Session` objects are documented as not thread-safe. Passing `db` to an executor thread while the calling coroutine (or other coroutines sharing the same session) may be mid-operation risks connection corruption, `DetachedInstanceError`, or silent data loss. This is the same bug flagged in PR #94 (KB retrieval service).

**Solution:** open a fresh `Session` inside the blocking function instead of sharing the caller's session:
```python
from app.db.session import SessionLocal

def _fetch_recent_summaries(
    tenant_id: uuid.UUID,
    call_flow_id: uuid.UUID,
    from_number: str,
    current_session_id: uuid.UUID,
    window: int,
) -> List[CallSession]:
    """Blocking DB query — runs in a thread pool with its own session."""
    db = SessionLocal()
    try:
        stmt = (
            select(CallSession)
            .where(
                CallSession.tenant_id == tenant_id,
                CallSession.call_flow_id == call_flow_id,
                CallSession.from_number == from_number,
                CallSession.status == "completed",
                CallSession.id != current_session_id,
                CallSession.transcript_summary.isnot(None),
                CallSession.transcript_summary != "",
            )
            .order_by(CallSession.start_time.desc())
            .limit(window)
        )
        return list(db.execute(stmt).scalars().all())
    finally:
        db.close()
```
Then update the call site to pass plain scalar arguments, not the ORM objects:
```python
sessions = await asyncio.wait_for(
    loop.run_in_executor(
        None,
        _fetch_recent_summaries,
        call_session.tenant_id,
        call_session.call_flow_id,
        call_session.from_number,
        call_session.id,
        window,
    ),
    timeout=fetch_timeout,
)
```

app/services/caller_memory_service.py
```diff
@@ -0,0 +1,123 @@
...
+def _format_caller_memory_block(sessions: List[CallSession]) -> str:
+    if not sessions:
+        return ""
+
+    lines = [f"CALLER HISTORY (last {len(sessions)} interactions):"]
+    for session in sessions:
+        date_str = (
+            session.start_time.strftime("%Y-%m-%d") if session.start_time else "unknown date"
+        )
+        summary = (session.transcript_summary or "").strip()
```
codewithwaleed: **`transcript_summary` injected into the system prompt without sanitization — second-order prompt injection.** The summary is generated by Gemini from what the caller said on a *previous* call. A malicious caller can deliberately speak prompt-injection text (e.g. `"Ignore your previous instructions and instead..."`) during a call; Gemini's summary may preserve enough of this phrasing to be injected into the agent's system prompt on the caller's next call (or if `from_number` is spoofed, into another tenant's call).

`_format_caller_memory_block` assembles summaries directly with no length cap or character sanitization:
```python
lines.append(f"- Call on {date_str}: {summary}")  # raw summary, no sanitization
```

**Fix:** sanitize each summary before injection and cap its length:
```python
import re

_CONTROL_CHARS_RE = re.compile(r'[\x00-\x1f\x7f]')
_MAX_SUMMARY_LEN = 400

def _sanitize_summary(text: str) -> str:
    text = _CONTROL_CHARS_RE.sub(' ', text).strip()
    if len(text) > _MAX_SUMMARY_LEN:
        text = text[:_MAX_SUMMARY_LEN] + '…'
    return text

def _format_caller_memory_block(sessions: List[CallSession]) -> str:
    if not sessions:
        return ""
    lines = [f"CALLER HISTORY (last {len(sessions)} interactions):"]
    for session in sessions:
        date_str = session.start_time.strftime("%Y-%m-%d") if session.start_time else "unknown date"
        summary = _sanitize_summary((session.transcript_summary or "").strip())
        lines.append(f"- Call on {date_str}: {summary}")
    lines.append("End of caller history.")
    return "\n".join(lines)
```
For stronger protection, consider wrapping the history block in clear delimiters (e.g. `<caller_history>...</caller_history>`) so the model has a clear signal about what is system instructions vs. data.

app/services/caller_memory_service.py
```diff
@@ -0,0 +1,123 @@
...
+            timeout=float(
+                getattr(settings, "VOICE_CALLER_MEMORY_FETCH_TIMEOUT_SEC", None)
+                or _DEFAULT_FETCH_TIMEOUT_SEC
+            ),
+        )
+        context_block = _format_caller_memory_block(sessions)
+    except asyncio.TimeoutError:
+        logger.warning(
+            "caller_memory lookup timed out after %sms; proceeding without caller memory",
+            int(_DEFAULT_FETCH_TIMEOUT_SEC * 1000),
```
codewithwaleed: **Timeout warning logs the module-level default constant instead of the actual timeout that was used.** If an operator sets `VOICE_CALLER_MEMORY_FETCH_TIMEOUT_SEC` in the environment, the actual timeout is that value — but the warning always says `100ms` (from `_DEFAULT_FETCH_TIMEOUT_SEC * 1000`) regardless. Ops investigating flapping caller-memory will see misleading log messages.

**Fix:** capture the resolved timeout into a local variable before the `try` block and reference it in the warning:
```python
fetch_timeout = float(
    getattr(settings, "VOICE_CALLER_MEMORY_FETCH_TIMEOUT_SEC", None)
    or _DEFAULT_FETCH_TIMEOUT_SEC
)
try:
    sessions = await asyncio.wait_for(
        loop.run_in_executor(...),
        timeout=fetch_timeout,
    )
    ...
except asyncio.TimeoutError:
    logger.warning(
        "caller_memory lookup timed out after %dms; proceeding without caller memory",
        int(fetch_timeout * 1000),  # ← actual timeout, not the constant
    )
```

codewithwaleed (changes requested):
Cross-session caller memory is a useful feature and the overall design is good — the fail-open timeout pattern, `begin_nested()` + `flush()` for metadata caching, and the composite index for the lookup query are all correct. Two issues need fixing before merge: a thread-safety violation that can corrupt the SQLAlchemy session, and an unsanitized transcript summary injected into the LLM system prompt.

PR #207, opened 2026-08-14
feat(voice): add AssemblyAI Universal-Streaming STT provider

Aliict:
## Summary
- Adds AssemblyAI Universal-Streaming as a new STT provider in `SttPipeline`, alongside Deepgram/Google/Speechmatics/ElevenLabs/xAI, following the same duck-typed session architecture.
- Native server-side turn detection only (`min_turn_silence`, `max_turn_silence`, `end_of_turn_confidence_threshold`, `vad_threshold`) — no local VAD. Twilio MULAW and LiveKit LINEAR16 audio sent directly, no conversion needed.
- Two-state `end_of_turn` transcript mapping (interim/final) — a deliberate contrast with xAI's three-state `is_final`/`speech_final` model, verified against AssemblyAI's official docs.
- No documented in-band error event for this API (unlike xAI): a connection loss before the server's `Termination` message is treated as fatal, and unrecognized server message types are now captured/logged rather than silently dropped (fixed after a live incident where the real error detail was lost — see commit `7a4b202`).
- Catalog seeded with a single model, `universal-streaming-multilingual`, per product decision (AssemblyAI also offers `universal-3-5-pro` and `universal-streaming-english`, intentionally not seeded).
- `ASSEMBLYAI_API_KEY` added to `.env.example`; production loads it via the existing AWS Secrets Manager → env-var mechanism, same as the other STT providers — no new mechanism.

## Test plan
- [x] `pytest tests/services/test_assemblyai_stt.py -v` — 25 passed
- [x] Full voice/STT regression suite (`tests/voice/` + all provider test files) — 476 passed, 0 regressions
- [x] `ruff check` clean on all changed files
- [x] Migration applied to dev DB (`alembic upgrade head`), seeded row verified by direct query
- [x] Live integration test (`RUN_ASSEMBLYAI_STT_INTEGRATION=1`) — gated, requires a real `ASSEMBLYAI_API_KEY`, not run in CI

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -877,12 +877,17 @@ class Settings(BaseSettings):
     # LiveKit/browser-call path (kb_retrieval_service via ConversationOrchestrator)
     # budget — distinct from RAG_RETRIEVAL_TIMEOUT_SEC (Twilio's bidirectional_stream
     # path) so tuning one never silently changes the other's latency contract.
-    # Observed embedding+vector-search samples were 271-450ms; 450ms left ~0ms
-    # margin and caused frequent false-positive timeouts (falling back to "no KB
-    # context" even when a result was about to arrive). 700ms keeps meaningful
-    # headroom over the observed p_max while staying well under ~1s, the point at
-    # which voice-call silence becomes perceptible to callers.
-    RAG_KB_RETRIEVAL_TIMEOUT_SEC: float = 0.7
+    # Originally set to 0.7s from local/CI samples of 271-450ms. Production
+    # diagnostics on the live EC2 deployment (2026-08-16) showed this environment's
+    # actual baseline round-trip + inference time to OpenAI's /v1/embeddings
+    # endpoint is itself 270-510ms with real variance (confirmed via raw httpx
+    # calls bypassing app code entirely), leaving ~0ms margin at 0.7s and causing
+    # frequent false-positive timeouts — KB context dropped even when a result was
+    # about to arrive. Raised to 1.2s to give this real-world baseline enough
+    # headroom while staying under the ~1.5-2s range where voice-call silence
+    # becomes clearly perceptible to callers. Revisit if p95 KB latency data
+    # (once available) suggests a tighter or looser value.
+    RAG_KB_RETRIEVAL_TIMEOUT_SEC: float = 1.2
     # Slow-path budget: cap cumulative waits for RAG/KB on a turn.
     VOICE_SLOWPATH_BUDGET_SEC: float = 0.55
     # How long we wait for an in-flight RAG prefetch before failing open.
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -865,7 +865,7 @@ class Settings(BaseSettings):
     # Similarity floor (0-1, higher = more similar) shared by both the Twilio path
     # (rag_context.py) and the LiveKit path (kb_retrieval_service._query_single_kb,
     # which returns 1 - cosine_distance as `score` — same direction/semantics).
-    RAG_SCORE_THRESHOLD: float = 0.4
+    RAG_SCORE_THRESHOLD: float = 0.2
     # Hard cap for the size of the rendered context block injected into prompts.
     # This is character-based (approx). For token-accurate sizing, you would need a tokenizer.
     RAG_MAX_CONTEXT_CHARS: int = 6000
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -604,7 +604,10 @@ class Settings(BaseSettings):
     # universal-streaming-english, deliberately not offered here).
     ASSEMBLYAI_STT_SPEECH_MODEL: str = "universal-streaming-multilingual"
     # Silence (ms) before AssemblyAI's native end-of-turn detection considers
-    # a turn boundary. Range 50-10000ms per docs; no vendor-documented default.
+    # a turn boundary. Range 50-10000ms per docs; no vendor-documented default
+    # -- 400ms chosen empirically as a conservative responsiveness/false-positive
+    # tradeoff, not a measured value. Revisit if end-of-turn detection feels
+    # too eager or too laggy in production.
     ASSEMBLYAI_STT_MIN_TURN_SILENCE_MS: int = 400
     # Vendor default per docs; same 50-10000ms range as min_turn_silence.
     ASSEMBLYAI_STT_MAX_TURN_SILENCE_MS: int = 1536
@@ -865,6 +868,11 @@ class Settings(BaseSettings):
     # Similarity floor (0-1, higher = more similar) shared by both the Twilio path
     # (rag_context.py) and the LiveKit path (kb_retrieval_service._query_single_kb,
     # which returns 1 - cosine_distance as `score` — same direction/semantics).
+    # Lowered from 0.4 -> 0.2 based on production telemetry showing relevant KB
+    # matches were being filtered out at 0.4. 0.2 is a deliberately low floor --
+    # verify with recall/precision metrics before lowering further. Irrelevant
+    # results at this threshold risk degrading LLM answer quality (hallucination
+    # risk) more than returning no KB context would.
     RAG_SCORE_THRESHOLD: float = 0.2
     # Hard cap for the size of the rendered context block injected into prompts.
     # This is character-based (approx). For token-accurate sizing, you would need a tokenizer.
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -161,6 +161,8 @@ class TtsSettings(BaseModel):
     provider: str = Field(default="elevenlabs", validation_alias="TTS_PROVIDER")
     api_key: str = Field(default="", validation_alias="TTS_API_KEY")
     rime_api_key: str = Field(default="", validation_alias="RIME_API_KEY")
+    hume_api_key: str = Field(default="", validation_alias="HUME_API_KEY")
+    hume_sample_rate_hz: int = Field(default=48000, validation_alias="HUME_TTS_SAMPLE_RATE_HZ")
     elevenlabs_api_key: str = Field(default="", validation_alias="ELEVENLABS_API_KEY")
     elevenlabs_encryption_key: str = Field(
         default="", validation_alias="ELEVENLABS_ENCRYPTION_KEY"
@@ -464,6 +466,17 @@ class Settings(BaseSettings):
 
     # Rime Labs TTS Configuration
     RIME_API_KEY: str = ""
+
+    # Hume AI TTS Configuration
+    HUME_API_KEY: str = ""
+    # ASSUMED default — Hume's PCM streaming output sample rate is not
+    # documented publicly as of this integration (2026-08). 48000 Hz is a
+    # reasonable industry-standard guess for a modern neural TTS PCM stream,
+    # but MUST be verified against a real Hume account/API key before this
+    # goes to production traffic; override via env var if Hume support
+    # confirms a different value. Drives PCMStreamDownsampler's src_rate_hz
+    # when converting Hume's PCM16LE chunks down to mulaw 8kHz for Twilio.
+    HUME_TTS_SAMPLE_RATE_HZ: int = 48000
 
     # ElevenLabs Configuration
     ELEVENLABS_API_KEY: str = ""
@@ -1114,6 +1127,8 @@ class Settings(BaseSettings):
             provider=self.TTS_PROVIDER,
             api_key=self.TTS_API_KEY,
             rime_api_key=self.RIME_API_KEY,
+            hume_api_key=self.HUME_API_KEY,
+            hume_sample_rate_hz=self.HUME_TTS_SAMPLE_RATE_HZ,
             elevenlabs_api_key=self.ELEVENLABS_API_KEY,
             elevenlabs_encryption_key=self.ELEVENLABS_ENCRYPTION_KEY,
             enable_audio_tags=self.ENABLE_ELEVENLABS_AUDIO_TAGS,
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -599,18 +599,20 @@ class Settings(BaseSettings):
     # the speech_model choice (english vs multilingual variants). Kept only
     # for interface parity/logging, same rationale as xAI's unused `model`.
     ASSEMBLYAI_STT_LANGUAGE: str = "en"
-    ASSEMBLYAI_STT_SPEECH_MODEL: str = "universal-3-5-pro"
+    # Only universal-streaming-multilingual is seeded in the sttmodel catalog
+    # (product decision -- AssemblyAI also offers universal-3-5-pro and
+    # universal-streaming-english, deliberately not offered here).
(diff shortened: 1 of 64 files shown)


app/services/assemblyai_stt_service.py
```diff
@@ -0,0 +1,446 @@
...
+                    logger.warning("[AssemblyAI STT] close failed: %s", exc)
+            except asyncio.CancelledError:
+                raise
+            except Exception as exc:  # noqa: BLE001
+                session_end_reason = "stream_exception"
+                logger.error("[AssemblyAI STT] sender loop error: %s", exc, exc_info=True)
+                await self._results_q.put(
+                    {"error": str(exc), "transcript": "", "confidence": 0.0, "is_final": True, "recoverable": False}
+                )
+                self._closed = True
```
codewithwaleed: 🔴 **Blocker: `self._closed` not set on normal `_sender_loop` exit — stale `push_audio` calls queue unboundedly**

When `_sender_loop` exits normally (graceful `Terminate` → `Termination` sequence), `self._closed` remains `False` at the `finally` block's entry. Any straggler `push_audio()` call that arrives after `_sender_loop` has exited still puts bytes into `_audio_q` — with the sender task gone, nothing ever drains it.

```python
finally:
    self._closed = True  # ← add this before push_done
    logger.info("[AssemblyAI STT] session_end reason=%s", session_end_reason)
    self._push_done()
    if self._receiver_task and not self._receiver_task.done():
        self._receiver_task.cancel()
```

app/services/assemblyai_stt_service.py
```diff
@@ -0,0 +1,446 @@
...
+            if not transcript:
+                # Empty/whitespace transcript must never reach the pipeline,
+                # regardless of end_of_turn.
+                return
+            end_of_turn = bool(data.get("end_of_turn"))
+            # Two-state only: end_of_turn=false -> interim, end_of_turn=true
+            # -> this Turn's transcript IS the finalized text for the turn.
+            # No third "chunk-final but not a boundary" state exists here.
+            self._results_q.put_nowait(
+                {"transcript": transcript, "confidence": 0.0, "is_final": end_of_turn}
```
codewithwaleed: 🟠 **Major: `confidence` hardcoded to `0.0` — bypasses all barge-in confidence gates**

AssemblyAI's `Turn` payload includes a `confidence` field. Setting `0.0` unconditionally means `_maybe_process_interim` and barge-in thresholds (`VOICE_BARGE_IN_MIN_CONFIDENCE`) always see zero — barge-in fires (or doesn't) based solely on word count, ignoring confidence entirely.

```python
def _handle_turn(self, data: Dict[str, Any]) -> None:
    transcript = (data.get("transcript") or "").strip()
    if not transcript:
        return
    end_of_turn = bool(data.get("end_of_turn"))
    confidence = float(data.get("confidence") or 0.0)  # ← use real value
    self._results_q.put_nowait(
        {"transcript": transcript, "confidence": confidence, "is_final": end_of_turn}
    )
```

app/voice/stt_pipeline.py
```diff
@@ -50,6 +52,8 @@ class SttPipeline:
                        native VAD/commit_strategy drives turn-end)
       "xai"          — XaiGrokSTTService (MULAW 8kHz, Twilio path; native
                        Smart Turn / endpointing drives turn-end)
+      "assemblyai"   — AssemblyAiSTTService (MULAW 8kHz Twilio / LINEAR16
+                       16kHz LiveKit path; native end_of_turn drives turn-end)
     """
```
codewithwaleed: 🟠 **Major: `asyncio.create_task(self._stt_session.start())` — untracked fire-and-forget**

If `start()` raises (e.g. WebSocket connect failure), the exception is stored in the abandoned task and never retrieved. Python emits a `Future exception was never retrieved` warning at GC time, but `_reader_loop` is already running and will block indefinitely on `get_result()`.

Keep a reference on the instance:
```python
self._reader_task = asyncio.create_task(self._reader_loop())
self._start_task = asyncio.create_task(self._stt_session.start())  # keep reference
```
(Same issue exists in `_ensure_xai_session` above — worth fixing there too.)

app/core/config.py
```diff
@@ -592,6 +592,30 @@ class Settings(BaseSettings):
...
+    # English/Multilingual model family (this deployment's only model).
+    ASSEMBLYAI_STT_END_OF_TURN_CONFIDENCE_THRESHOLD: float = 0.4
+    # Vendor default for the Universal-family models per docs. Range 0.0-1.0.
+    ASSEMBLYAI_STT_VAD_THRESHOLD: float = 0.4
+    # Whether finalized Turn transcripts are punctuation/cased-formatted.
+    ASSEMBLYAI_STT_FORMAT_TURNS: bool = True
+
     # Deepgram Speech-to-Text (replaces Google STT for streaming + batch)
     DEEPGRAM_API_KEY: str = ""
     DEEPGRAM_STT_MODEL: str = "nova-3"
```
codewithwaleed: 🔴 **Blocker: `RAG_SCORE_THRESHOLD` lowered from 0.4 → 0.2 without a precision guard — may inject irrelevant KB content into all calls**

0.2 cosine similarity is quite low — documents sharing only a few common words will match. On a live voice call, injecting irrelevant KB content can cause the LLM to confidently state incorrect information, which is worse than injecting none.

This change needs either:
1. A minimum-score or minimum-result-count guard in the retrieval path, or
2. A staged rollout (opt-in per tenant), or
3. A documented decision with telemetry showing recall/precision at 0.2 vs 0.4

At minimum, add a comment noting the risk so a future maintainer doesn't lower it further without thought:
```python
# NOTE: 0.2 is a deliberately low floor — verify with recall/precision metrics
# before lowering further. Irrelevant results at this threshold may degrade
# LLM answer quality (hallucination risk) more than no KB context would.
RAG_SCORE_THRESHOLD: float = 0.2
```

app/services/assemblyai_stt_service.py
```diff
@@ -0,0 +1,446 @@
...
+    return value
+
+
+def _mask_key(key: str | None) -> str:
+    """Safe, non-reversible preview for diagnostics — never logs the raw key."""
+    if not key:
+        return "<empty>"
+    if len(key) <= 8:
+        return f"<masked len={len(key)}>"
+    return f"{key[:4]}...{key[-2:]} (len={len(key)})"
```
codewithwaleed: 🟡 **Minor: `_mask_key` leaks 6 characters of the API key (`key[:4]` + `key[-2:]`)**

For a short key (10-12 chars), showing 4 + 2 = 6 characters is a significant fraction. Other secret-manager preview functions show only a fixed 4-char prefix. Use:

```python
def _mask_key(key: str | None) -> str:
    if not key:
        return "<empty>"
    if len(key) <= 8:
        return f"<masked len={len(key)}>"
    return f"{key[:4]}{'*' * min(8, len(key) - 4)} (len={len(key)})"
```

app/core/config.py
```diff
@@ -592,6 +592,30 @@ class Settings(BaseSettings):
...
+    # the speech_model choice (english vs multilingual variants). Kept only
+    # for interface parity/logging, same rationale as xAI's unused `model`.
+    ASSEMBLYAI_STT_LANGUAGE: str = "en"
+    # Only universal-streaming-multilingual is seeded in the sttmodel catalog
+    # (product decision -- AssemblyAI also offers universal-3-5-pro and
+    # universal-streaming-english, deliberately not offered here).
+    ASSEMBLYAI_STT_SPEECH_MODEL: str = "universal-streaming-multilingual"
+    # Silence (ms) before AssemblyAI's native end-of-turn detection considers
+    # a turn boundary. Range 50-10000ms per docs; no vendor-documented default.
+    ASSEMBLYAI_STT_MIN_TURN_SILENCE_MS: int = 400
```
codewithwaleed: 🟡 **Minor: `ASSEMBLYAI_STT_MIN_TURN_SILENCE_MS=400` has no documented justification**

The adjacent `ASSEMBLYAI_STT_MAX_TURN_SILENCE_MS=1536` cites "vendor default per docs", but `min_turn_silence_ms=400` has no such attribution. If this was chosen empirically or is just a conservative guess, document why — it affects end-of-turn detection latency directly and future maintainers need to know whether it's a measured value or an arbitrary default:

```python
# 400ms — chosen empirically; no vendor-published default for this param.
# Lower values increase responsiveness but risk false end-of-turn on brief pauses.
ASSEMBLYAI_STT_MIN_TURN_SILENCE_MS: int = 400
```

codewithwaleed (changes requested):
## PR #207 Review — Add AssemblyAI Universal-Streaming STT provider

### Summary
This PR integrates AssemblyAI's Universal-Streaming STT via raw WebSocket, following the established adapter pattern from the xAI/Speechmatics/ElevenLabs integrations. The two-state `Turn` message model (no intermediate "chunk-final" state), the `Terminate` → `Termination` graceful shutdown sequence, and the `_ready_evt` gate before sending audio are all correctly implemented and well-documented. The accompanying config changes (raising `RAG_KB_RETRIEVAL_TIMEOUT_SEC` from 0.7s to 1.2s with production telemetry justification, and lowering `RAG_SCORE_THRESHOLD` from 0.4 to 0.2) are well-motivated.

A few correctness issues need attention before merge.

---

### 🔴 Blockers

**1. `create_streaming_session` raises `RuntimeError` when `self._api_key` is falsy, but `AssemblyAiSTTService.__init__` sets `self._api_key = None` (not `""`) when the key is empty**

```python
# __init__
key = (settings.ASSEMBLYAI_API_KEY or "").strip()
self._api_key = key or None   # ← None if empty

# create_streaming_session
if not self._api_key:
    raise RuntimeError("AssemblyAI client not initialized — set ASSEMBLYAI_API_KEY")
```

`not None` is `True`, so the guard fires correctly. This is fine as written — no bug. ✓

**2. `_sender_loop`'s `finally` block calls `self._push_done()` unconditionally, but `_push_done` is also called in `_receiver_loop`'s `finally` — `_done_pushed` guards against double-push, but `_closed = True` is set only inside `_sender_loop`'s except branches, not on the normal path**

When `_sender_loop` exits normally (client sent `finish()`, Terminate was acknowledged, socket closed), `self._closed` is still `False` at the `finally` block's entry point. That means a subsequent `push_audio()` call (from a straggler audio chunk in the `_audio_q`) would still call `self._audio_q.put_nowait(chunk)` even though the WebSocket is already closed — the `_sender_task` is gone and nothing drains that queue anymore. The queue grows unboundedly.

Fix — set `self._closed = True` at the top of `_sender_loop`'s `finally` block, before `_push_done`:

```python
finally:
    self._closed = True   # ← add this line
    logger.info("[AssemblyAI STT] session_end reason=%s", session_end_reason)
    self._push_done()
    if self._receiver_task and not self._receiver_task.done():
        self._receiver_task.cancel()
```

**3. `RAG_SCORE_THRESHOLD` lowered from 0.4 to 0.2 without a minimum result score or count guard — could inject low-relevance KB content into every call**

The threshold change is driven by valid production observations, but 0.2 cosine similarity is quite low for RAG — at this threshold, documents that share only a few common words with the query will match. For a live voice call, injecting irrelevant KB content into the prompt is potentially worse than injecting none (it can cause the LLM to confidently state incorrect information). This threshold change deserves a staged rollout or a minimum-score flag in the retrieval path, not a blanket production change. Consider documenting the decision and adding a note that this should be revisited once recall/precision data is available from production.

---

### 🟠 Major Issues

**4. `_ensure_assemblyai_session` uses `asyncio.create_task(self._stt_session.start())` — the start task is never tracked, so an error in `start()` is silently dropped**

```python
self._reader_task = asyncio.create_task(self._reader_loop())
asyncio.create_task(self._stt_session.start())  # ← no reference kept
```

If `start()` raises (e.g. WebSocket connect failure), the exception is stored in the task but never retrieved — Python will emit an `asyncio: Future exception was never retrieved` warning at GC time, but the failure is invisible to the caller. Meanwhile `_reader_loop` is already running and will wait indefinitely on `get_result()`.

This is the same pattern as `_ensure_xai_session` (which has the same issue), but it's still worth fixing here. Assign the task and add a done callback:

```python
start_task = asyncio.create_task(self._stt_session.start())
# Tasks in start() push errors to results_q — _reader_loop already drains that.
# Keep a reference so GC doesn't collect it prematurely.
self._start_task = start_task
```

**5. The `confidence` field in every `_handle_turn` result is hardcoded to `0.0`**

```python
self._results_q.put_nowait(
    {"transcript": transcript, "confidence": 0.0, "is_final": end_of_turn}
)
```

AssemblyAI's `Turn` payload includes a `confidence` field. Hardcoding `0.0` means the barge-in / quick-ack confidence thresholds in `ConversationOrchestrator` / `_maybe_process_interim` will always see `0.0` confidence — the barge-in gate (`VOICE_BARGE_IN_MIN_CONFIDENCE`) fires based on confidence, so this effectively bypasses the confidence check for all AssemblyAI interims.

Fix:
```python
confidence = float(data.get("confidence") or 0.0)
self._results_q.put_nowait(
    {"transcript": transcript, "confidence": confidence, "is_final": end_of_turn}
)
```

---

### 🟡 Minor Issues

**6. `ASSEMBLYAI_STT_MIN_TURN_SILENCE_MS` documented as 400ms default — but no vendor-documented default is mentioned for this param, unlike `max_turn_silence` (1536ms) which cites the vendor default**

The comment says "no vendor-documented default" for `min_turn_silence`, but then sets 400ms as the value with no further justification. If this was chosen empirically, document why 400ms was chosen so a future maintainer knows whether it's a conservative guess or a measured value.

**7. `_mask_key` returns `f"{key[:4]}...{key[-2:]} (len={len(key)})"` — leaks 6 characters of the API key in logs**

For short keys (8 < len ≤ 12), 4+2=6 characters is a significant fraction of the key. The existing pattern in other secret-manager functions uses a fixed 4-char preview only. Use `key[:4]` + masked suffix:

```python
return f"{key[:4]}{'*' * (len(key) - 4)} (len={len(key)})"
```

**8. The module-level `assemblyai_stt_service = AssemblyAiSTTService()` singleton is constructed at import time**

Unlike `HumeTtsService` (PR #212 blocker), `AssemblyAiSTTService.__init__` only reads settings and logs a warning — it does NOT raise on a missing key, so this is not a blocker. It is still consistent with the pattern of the other STT services (which do the same). ✓

---

### Missing Test Coverage

- **`_handle_turn` with a non-zero confidence value in the payload**: the current hardcoded `0.0` issue (#5) would be caught by a test that asserts `result["confidence"] == data["confidence"]`.
- **`_sender_loop` `finally` with `self._closed` check**: a test that calls `push_audio` after `finish()` has been called (and `_sender_loop` has exited) would confirm the queue doesn't grow unboundedly.
- **`RAG_SCORE_THRESHOLD=0.2` integration test**: a test with a query that has 0.25 cosine similarity should now be included to confirm the new threshold doesn't over-include irrelevant results.

---

### Recommendation: **Request Changes** — fix the `confidence` hardcode (#5) and the `_closed` not being set on normal sender exit (#2) before merge.

PR #208, opened 2026-08-16
GPT-5 model support, OpenAI/Groq async streaming fix, LiveKit RAG prefetch, KB retrieval latency fixes

Aliict:
## Summary

- **OpenAI/Groq async streaming fix**: `stream_text` on both services was iterating a synchronous SDK client inside `async def`, blocking the event loop for all concurrent calls during token generation. Switched to a cached `AsyncOpenAI`/`AsyncAzureOpenAI` client with genuine `async for` streaming.
- **New OpenAI GPT models**: added `gpt-5`, `gpt-5-mini`, `gpt-5.1`, `gpt-5.2`, `gpt-5.4` (model IDs verified against official OpenAI docs). GPT-5-family requests use `max_completion_tokens` (floored at 1500 to avoid the reasoning-token budget silently starving visible output) instead of `max_tokens`, and omit `temperature`. Added an additive-only migration widening the `ck_agent_llm_model` CHECK constraint to match.
- **Explicit `reasoning_effort` for GPT-5-family models**: `gpt-5`/`gpt-5-mini` were silently running OpenAI's implicit `medium` reasoning effort on every voice-agent turn (confirmed via the installed SDK's own parameter docs). Now set to `low`; `gpt-5.1`/`gpt-5.2`/`gpt-5.4` set to `none` explicitly. This is a plausible mechanism for both elevated GPT-5 latency and unreliable KB grounding reported in production. Non-reasoning OpenAI models are untouched.
- **LiveKit/browser RAG interim prefetch**: ported the Twilio RAG interim-prefetch pattern (fire on qualifying STT interim, consume once on final, cancel on barge-in) to the browser call path, overlapping KB retrieval latency with remaining user speech. Includes a staleness guard (`rag_prefetch_matches_final`) and a barge-in-ordering fix so prefetch is never wastefully fired-then-cancelled on interrupted turns.
- **KB retrieval cold-start latency fix**: `embedding_service` was constructing a brand-new OpenAI client (and connection pool) on every embedding call, not just the first. Now reuses the existing cached-client pattern, plus a non-blocking, per-worker startup warm-up (staging/production only).
- **`RAG_KB_RETRIEVAL_TIMEOUT_SEC` raised 0.7s → 1.2s** and **`RAG_SCORE_THRESHOLD` lowered 0.4 → 0.2**, both based on live production diagnostics (raw embedding latency and real similarity-score measurements against a production KB, documented in commit messages/comments).

## Verification

- Twilio path (`bidirectional_stream.py`) confirmed zero-diff throughout — untouched.
- Vertex Gemini, existing OpenAI models, Groq (non-gpt-5), STT/TTS providers, Calendly, CRM context unaffected.
- Full test suite passing (2472+ tests), ruff clean on all changed files.
- Independent code-review passes run after each change (async client caching, RAG prefetch ordering, reasoning_effort config) — findings addressed before merge.

## Known residual risks (see commit messages for detail)

- `gpt-5.2`/`gpt-5.4`'s `reasoning_effort="none"` default is extrapolated from `gpt-5.1`'s documented behavior, not individually confirmed — worth a live smoke test once API credits allow.
- KB content-quality issue found during production diagnosis (duplicated chunks, raw JSON fragments instead of natural-language chunking) — threshold lowered as an immediate mitigation, but re-ingesting the affected KB with cleaner chunking would likely improve match quality further. Not addressed in this PR (config-only fix).

### DIFF
--- a/app/services/embedding_service.py
+++ b/app/services/embedding_service.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 import time
 
 from app.core.config import settings
@@ -86,8 +85,20 @@ async def warm_embedding_client() -> None:
         return
     t0 = time.perf_counter()
     try:
-        loop = asyncio.get_running_loop()
-        await loop.run_in_executor(None, embed_text_for_rag, "warmup")
+        # Use the async client directly rather than running the sync
+        # embed_text_for_rag() in a thread-pool executor: openai_service's
+        # cached sync client wraps a single httpx.Client, which is not
+        # thread-safe for concurrent use. The warm-up itself is a single
+        # one-shot call so it was benign today, but routing it through
+        # run_in_executor established a pattern that would silently become
+        # unsafe if this ever ran concurrently with other executor-thread
+        # embedding calls. The async client has no such constraint.
+        from app.services.openai_service import openai_service
+
+        client = openai_service.get_async_client()
+        await client.embeddings.create(
+            model=settings.RAG_EMBEDDING_MODEL, input="warmup"
+        )
         logger.info(
             "kb_retrieval embedding client warm-up complete in %.1fms",
             (time.perf_counter() - t0) * 1000,
--- a/app/services/embedding_service.py
+++ b/app/services/embedding_service.py
@@ -1,6 +1,10 @@
 from __future__ import annotations
 
+import asyncio
+import time
+
 from app.core.config import settings
+from app.core.logger import logger
 
 
 def embed_text_for_rag(text: str) -> list[float]:
@@ -41,11 +45,54 @@ def embed_text_for_rag(text: str) -> list[float]:
 
 
 def _embed_openai_ada002(text: str) -> list[float]:
-    from app.core.openai_client import get_openai_client
+    # Route through openai_service's cached-by-api-key client dict (see
+    # OpenAIService.get_client) instead of app.core.openai_client.get_openai_client()
+    # directly. get_openai_client() constructs a brand-new OpenAI(...) client
+    # (and therefore a brand-new httpx.Client / connection pool) on every call —
+    # previously this function called it fresh on every single embedding, so no
+    # TCP/TLS connection was ever kept warm across calls, and the very first
+    # embedding call in a freshly started worker process additionally pays for
+    # one-time httpx/certifi import + CA-bundle parsing + cold DNS resolution on
+    # top of the handshake, which is what pushed the first-in-process retrieval
+    # past RAG_KB_RETRIEVAL_TIMEOUT_SEC in production while every later, already-
+    # warm call landed around ~200ms. Reusing the cached client lets the
+    # underlying connection pool keep the TCP/TLS session alive across calls.
+    from app.services.openai_service import openai_service
 
-    client = get_openai_client()
+    client = openai_service.get_client()
     resp = client.embeddings.create(
         model=settings.RAG_EMBEDDING_MODEL,
         input=text,
     )
     return resp.data[0].embedding
+
+
+async def warm_embedding_client() -> None:
+    """
+    Fire a single, cheap embedding call to pre-establish the cached OpenAI
+    client's TCP/TLS connection (and pay any one-time module-import /
+    CA-bundle-parsing / DNS-resolution cost) before the first real call
+    arrives on this worker process.
+
+    Intended to be scheduled as a non-blocking, fire-and-forget background
+    task from app startup (`asyncio.create_task`, never awaited inline) — see
+    app/main.py's lifespan. Must never raise: this is best-effort latency
+    warm-up, not a startup dependency, so any failure (no API key configured,
+    transient network issue, etc.) is logged and swallowed rather than
+    surfaced. Each Uvicorn worker process runs its own lifespan, so this
+    naturally warms one client per worker rather than a single shared one.
+    """
+    if not settings.OPENAI_API_KEY:
+        return
+    t0 = time.perf_counter()
+    try:
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(None, embed_text_for_rag, "warmup")
+        logger.info(
+            "kb_retrieval embedding client warm-up complete in %.1fms",
+            (time.perf_counter() - t0) * 1000,
+        )
+    except Exception as exc:
+        logger.warning(
+            "kb_retrieval embedding client warm-up failed (non-fatal): %s", exc
+        )
--- a/app/services/groq_service.py
+++ b/app/services/groq_service.py
@@ -4,34 +4,57 @@ Handles all Groq-related operations including text generation
 Groq uses OpenAI-compatible API with ultra-fast inference
 """
 
-from openai import OpenAI
+from openai import AsyncOpenAI, OpenAI
 from app.core.config import settings
 from typing import List, Dict, Any
 import time
 
+_GROQ_BASE_URL = "https://api.groq.com/openai/v1"
+
 class GroqService:
     """Service class for handling Groq operations"""
-    
+
     def __init__(self):
         self._clients = {}  # Store clients by API key
+        self._async_clients = {}  # Store async clients by API key (streaming hot path)
         self._current_api_key = None
-    
+
     def get_client(self, api_key: str = None):
         """Get or create Groq client with specific API key"""
         # Use provided API key or fall back to global setting (if exists)
         key_to_use = api_key or getattr(settings, 'GROQ_API_KEY', None)
-        
+
         if not key_to_use:
             raise Exception("Groq API key not found. Please provide an API key from database (model.api_key).")
-        
+
         # Return existing client or create new one for this API key
         if key_to_use not in self._clients:
             self._clients[key_to_use] = OpenAI(
                 api_key=key_to_use,
-                base_url="https://api.groq.com/openai/v1"
+                base_url=_GROQ_BASE_URL
             )
-        
+
         return self._clients[key_to_use]
+
+    def get_async_client(self, api_key: str = None):
+        """Get or create an AsyncOpenAI (Groq-compatible) client for streaming.
+
+        Groq's SDK surface is OpenAI-compatible; used only by stream_text so
+        the event loop is never blocked by sync HTTP/SSE calls in the
+        real-time voice hot path.
+        """
+        key_to_use = api_key or getattr(settings, 'GROQ_API_KEY', None)
+
+        if not key_to_use:
+            raise Exception("Groq API key not found. Please provide an API key from database (model.api_key).")
+
+        if key_to_use not in self._async_clients:
+            self._async_clients[key_to_use] = AsyncOpenAI(
+                api_key=key_to_use,
+                base_url=_GROQ_BASE_URL
+            )
+
+        return self._async_clients[key_to_use]
     
     def generate_text(self, prompt: str, system_prompt: str = None, 
                      model_name: str = "llama-3.3-70b-versatile", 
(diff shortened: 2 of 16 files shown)


app/voice/conversation_orchestrator.py
```diff
@@ -501,25 +531,81 @@ async def generate_and_stream_response(
...
+            # to this transport's kb_retrieval_service call. Falls back to the
+            # exact same synchronous retrieval this code path always used
+            # when no prefetch fired (e.g. a very short first utterance) or
+            # when the final utterance diverged materially from the interim
+            # that triggered the prefetch.
             kb_context_block = ""
             flow = getattr(self._h, "call_flow", None)
             flow_kb_ids = (flow.knowledge_base_ids or []) if flow else []
             if flow_kb_ids and self._h.db:
-                try:
```
codewithwaleed: 🔴 **Blocker: `_prefetch.result()` can raise if the task stored an exception**

`_prefetch_kb_context` currently always catches and returns `("", 0.0)` so this is safe today — but the implicit contract is fragile. If a future change lets `_prefetch_kb_context` re-raise, this line will throw inside the `try` block and fall to the outer `except Exception` handler, silently discarding the error.

Add an explicit exception check:
```python
if _prefetch.done():
    exc = _prefetch.exception()
    if exc:
        logger.debug("[RAG prefetch] stored exception: %s", exc)
        kb_context_block, kb_latency_ms = "", 0.0
    else:
        kb_context_block, kb_latency_ms = _prefetch.result()
```

app/services/openai_service.py
```diff
@@ -4,31 +4,160 @@
...
+    """Build the model/temperature/max-tokens kwargs for a Chat Completions call.
+
+    Isolated here so all OpenAI call sites in this service (streaming and
+    non-streaming) apply the gpt-5.x parameter differences identically.
+    """
+    if _is_reasoning_model(model_name):
+        # temperature intentionally omitted: gpt-5.x rejects/ignores non-default
+        # temperature on Chat Completions.
+        floored_max_tokens = max(int(max_tokens or 0), _REASONING_MODEL_MIN_COMPLETION_TOKENS)
+        params: Dict[str, Any] = {
```
codewithwaleed: 🟠 **Major: `max_tokens=None` is silently coerced to 1500 for all gpt-5 callers**

`int(max_tokens or 0)` converts `None` → `0` → `max(..., 1500)`. This means callers that pass `max_tokens=None` (intending to omit the parameter and let the model decide) get a hard 1500-token ceiling instead. Non-voice callers may legitimately want an uncapped budget.

```python
# current
floored_max_tokens = max(int(max_tokens or 0), _REASONING_MODEL_MIN_COMPLETION_TOKENS)

# fix
if max_tokens is not None and int(max_tokens) > 0:
    floored_max_tokens = max(int(max_tokens), _REASONING_MODEL_MIN_COMPLETION_TOKENS)
else:
    floored_max_tokens = _REASONING_MODEL_MIN_COMPLETION_TOKENS
```

app/services/embedding_service.py
```diff
@@ -41,11 +45,54 @@ def embed_text_for_rag(text: str) -> list[float]:
 
 
 def _embed_openai_ada002(text: str) -> list[float]:
-    from app.core.openai_client import get_openai_client
+    # Route through openai_service's cached-by-api-key client dict (see
+    # OpenAIService.get_client) instead of app.core.openai_client.get_openai_client()
+    # directly. get_openai_client() constructs a brand-new OpenAI(...) client
+    # (and therefore a brand-new httpx.Client / connection pool) on every call —
+    # previously this function called it fresh on every single embedding, so no
+    # TCP/TLS connection was ever kept warm across calls, and the very first
```
codewithwaleed: 🟠 **Major: `run_in_executor` shares a sync `httpx.Client` across threads**

`embed_text_for_rag` → `_embed_openai_ada002` → `openai_service.get_client()` returns a sync `openai.OpenAI` whose underlying `httpx.Client` is **not thread-safe** for concurrent use. The warm-up fires once at startup so it's benign today, but the pattern is fragile if concurrency is ever introduced.

Prefer the async client directly — no thread pool needed:
```python
async def warm_embedding_client() -> None:
    if not settings.OPENAI_API_KEY:
        return
    t0 = time.perf_counter()
    try:
        from app.services.openai_service import openai_service
        client = openai_service.get_async_client()
        await client.embeddings.create(model=settings.RAG_EMBEDDING_MODEL, input="warmup")
        logger.info("embedding warm-up complete in %.1fms", (time.perf_counter() - t0) * 1000)
    except Exception as exc:
        logger.warning("embedding warm-up failed (non-fatal): %s", exc)
```

app/voice/conversation_orchestrator.py
```diff
@@ -121,6 +121,36 @@ def should_send_quick_ack(user_text: str, config: QuickAckConfig) -> bool:
...
+def rag_prefetch_matches_final(prefetch_source_text: str, final_text: str) -> bool:
+    """
+    Decide whether a RAG/KB prefetch fired on an STT *interim* is still safe
+    to reuse for the STT *final* of the same turn.
+
+    The Twilio path (bidirectional_stream.py's `_prefetch_rag_context`) fires
+    on the first qualifying interim and always reuses whatever result that
+    produced, with no divergence check — it accepts that a slight interim/
+    final wording difference is fine for retrieval purposes. The browser path
+    adds this lightweight safety check on top of that same pattern: if the
```
codewithwaleed: 🟡 **Minor: dead-code guard after `set.split()` — `not source_words` / `not final_words` is unreachable**

Both `source` and `final` are already confirmed non-empty by the guard two lines above, so `.split()` on a non-empty string can never return an empty list here. The second guard is unreachable and can be removed to reduce noise:
```python
source_words = set(source.split())
final_words = set(final.split())
overlap = len(source_words & final_words) / max(len(source_words), len(final_words))
return overlap >= 0.6
```

app/services/groq_service.py
```diff
@@ -111,28 +134,29 @@ async def stream_text(self, prompt: str, system_prompt: str = None,
...
             )
-            
-            for chunk in stream:
+
+            async for chunk in stream:
                 if chunk.choices[0].delta.content is not None:
                     yield chunk.choices[0].delta.content
-                    
+
         except Exception as e:
```
codewithwaleed: 🟡 **Minor: `raise Exception(...)` destroys original exception type in streaming hot path**

`openai.RateLimitError`, `openai.AuthenticationError`, etc. carry structured data (status code, headers, retry-after) that retry logic higher in the stack might want to inspect. Wrapping in a bare `Exception` discards all of that.

```python
# instead of:
except Exception as e:
    raise Exception(f"Error in Groq streaming: {str(e)}")

# prefer:
except Exception:
    raise  # or: raise RuntimeError("Groq streaming error") from e
```
Same pattern applies to `OpenAIService.stream_text`.

codewithwaleed (changes requested):
## PR #208 Review — GPT-5 support, async streaming fix, LiveKit RAG prefetch

### Summary
This PR ships four related improvements: (1) GPT-5 reasoning-model family support with correct `max_completion_tokens`/`reasoning_effort` routing, (2) the async streaming fix that eliminates event-loop blocking in the real-time voice hot path, (3) LiveKit browser-call RAG interim-prefetch (mirrors the existing Twilio path), and (4) embedding client warm-up at startup. The code is well-commented and architecturally consistent with the existing patterns. A few issues need attention before merge.

---

### 🔴 Blockers

**1. `asyncio.shield` swallows `TimeoutError` on a completed-but-raised task**
When the shielded prefetch task *already completed with an exception* before this timeout fires, `asyncio.shield` re-raises that stored exception inside `wait_for`, which then gets caught by the outer `except Exception` handler — silently. But if the task is *still running* and `wait_for` times out, `asyncio.shield` correctly keeps the inner task alive. The real bug is that `_prefetch.result()` on line `kb_context_block, kb_latency_ms = _prefetch.result()` (the "already finished — zero-cost read" branch) can raise if the prefetch task stored an exception. Since `_prefetch_kb_context` always catches and returns `("", 0.0)`, this is currently safe — but only by an implicit contract. Add an explicit guard so a future change to `_prefetch_kb_context` doesn't silently swallow a hard error.

Fix:
```python
if _prefetch.done():
    exc = _prefetch.exception()
    if exc:
        logger.debug("[RAG prefetch] prefetch task stored exception: %s", exc)
        kb_context_block, kb_latency_ms = "", 0.0
    else:
        kb_context_block, kb_latency_ms = _prefetch.result()
```

---

### 🟠 Major Issues

**2. `_completion_params` silently floors `max_tokens=0` → `1500` for all gpt-5 callers**
`max(int(max_tokens or 0), _REASONING_MODEL_MIN_COMPLETION_TOKENS)` means that when a caller explicitly passes `max_tokens=0` (meaning "let the model decide" — a legitimate value on some OpenAI endpoints) it gets silently rewritten to 1500. More importantly, `max_tokens or 0` coerces `None` to `0`, then floors to 1500 — so `max_tokens=None` (omit the param entirely, another legitimate intent) also becomes 1500 rather than being omitted. For non-voice callers that want to give the model a free budget, this forces a hard ceiling they didn't ask for.

Fix:
```python
if _is_reasoning_model(model_name):
    params: Dict[str, Any] = {"model": model_name}
    if max_tokens is not None and int(max_tokens) > 0:
        params["max_completion_tokens"] = max(int(max_tokens), _REASONING_MODEL_MIN_COMPLETION_TOKENS)
    else:
        params["max_completion_tokens"] = _REASONING_MODEL_MIN_COMPLETION_TOKENS
    effort = _REASONING_EFFORT_BY_MODEL.get((model_name or "").strip().lower())
    if effort:
        params["reasoning_effort"] = effort
    return params
```

**3. `warm_embedding_client` uses `run_in_executor` with the module-level `embed_text_for_rag`, which internally calls `openai_service.get_client()` — sharing the sync client across threads**
`embed_text_for_rag` → `_embed_openai_ada002` → `openai_service.get_client()` returns a *sync* `openai.OpenAI` client. That client owns a `httpx.Client` which is not thread-safe to use concurrently. The warm-up runs in a thread pool executor while the main event loop may also be dispatching embedding calls. In practice this is benign (the warm-up fires once at startup before real traffic), but the pattern is fragile — a future change that runs two concurrent warm-ups (e.g. multiple UV workers calling lifespan simultaneously on the same process) would hit a real race.

Fix — run the warm-up directly in the event loop using the *async* client that already exists:
```python
async def warm_embedding_client() -> None:
    if not settings.OPENAI_API_KEY:
        return
    t0 = time.perf_counter()
    try:
        from app.services.openai_service import openai_service
        async_client = openai_service.get_async_client()
        await async_client.embeddings.create(
            model=settings.RAG_EMBEDDING_MODEL,
            input="warmup",
        )
        logger.info("embedding client warm-up complete in %.1fms", (time.perf_counter() - t0) * 1000)
    except Exception as exc:
        logger.warning("embedding client warm-up failed (non-fatal): %s", exc)
```

---

### 🟡 Minor Issues

**4. `rag_prefetch_matches_final` returns `False` when both inputs are empty, but that case is already guarded**
The early `if not source or not final: return False` before the word-set computation is correct. However the second guard `if not source_words or not final_words: return False` after splitting can never be reached — `.split()` on an empty string already returns `[]`, but we just checked `not source` above. This is dead code that adds noise. Remove it:
```python
source_words = set(source.split())
final_words = set(final.split())
overlap = len(source_words & final_words) / max(len(source_words), len(final_words))
return overlap >= 0.6
```

**5. `stream_text` in both `OpenAIService` and `GroqService` wraps all exceptions in a bare `raise Exception(...)`**
This destroys the original exception type (`openai.RateLimitError`, `openai.AuthenticationError`, etc.) that callers higher in the stack might want to inspect for retry logic or user-facing error messages. The conversation orchestrator's `try_stream` catches `Exception` generically, so it works today — but the pattern is a footgun for future error-specific handling.

Fix:
```python
except Exception as e:
    raise  # or: raise RuntimeError(f"OpenAI streaming error") from e
```

**6. `_GROQ_BASE_URL` constant is defined between the import block and the class definition with no blank line after imports**
Minor PEP 8 / project style: a blank line is missing between `import time` and `_GROQ_BASE_URL = ...`. Not a correctness issue.

---

### 🔵 Nits

- `_completion_params` is a module-level function but it imports `Dict` from `typing` which is already imported at the top — consistent, no issue.
- The migration's `_llm_check_sql` uses string formatting (`%`) instead of parameterized queries, but since values are compile-time string literals from a controlled tuple this is fine.
- `_REASONING_EFFORT_BY_MODEL` lookup uses `.get(...)` which returns `None` for unknown models and skips `reasoning_effort` — the correct fallback. ✓
- `asyncio.shield` usage in `generate_and_stream_response` is correct and well-motivated by the comment. ✓

---

### Missing Test Coverage

- **`_completion_params` with `max_tokens=None` and `max_tokens=0`** — the existing tests exercise positive values; the edge cases above (issue #2) are not covered.
- **`rag_prefetch_matches_final` with empty strings** — the function has an early-return path for empty input; a test confirming `("", "hello") → False` and `("hello", "") → False` would catch regressions.
- **`warm_embedding_client` failure path** — the test file covers the happy path; a test that mocks `embed_text_for_rag` to raise and asserts no exception propagates would guard the fail-open contract.

---

### Recommendation: **Request Changes** (issues #2 and #3 before merge; #1 is low-risk but worth the one-line fix)

The async streaming fix and the GPT-5 routing logic are correct and well-motivated. The RAG prefetch port from Twilio to LiveKit is architecturally sound. The three issues above are fixable in this branch without restructuring anything.

PR #209, opened 2026-08-16
Migrate Vertex Gemini service to google-genai SDK; add Gemini 3 model support

Aliict:
## Summary

- Migrates `VertexGeminiService` off the deprecated `vertexai.generative_models` SDK (removal date already passed) onto the replacement `google-genai` SDK, preserving streaming, barge-in cancellation, and Calendly tool-calling behavior. Fixed a `role="function"` vs `role="user"` mismatch in the tool-calling follow-up turn during the migration (caught in code review, would have silently broken live appointment booking).
- Adds `gemini-3-flash-preview` and `gemini-3.1-flash-lite` to the agent `llm_model` allow-list (Python allow-list + DB CHECK constraint migration) and pricing table.
- Fixes a live 404: Google currently only serves the Gemini 3 family from Vertex AI's `global` endpoint, not the configured regional endpoint (`us-central1`). Added per-model region routing (`_location_for_model`) with a client cached per location — `gemini-2.5-flash` and all other models are unaffected and keep using the regional endpoint.
- Fixes a live empty-response bug: `gemini-3-flash-preview`'s default `thinking_level="high"` (and an initial `"low"` attempt) consumes tokens from the same `max_output_tokens` budget used for the spoken reply, and could exhaust it entirely (`finish_reason=MAX_TOKENS`, zero output). Pinned both Gemini 3 models to `thinking_level="minimal"`, verified live to reliably fit the answer within this service's existing token budgets.

## Operational note

`gemini-3-flash-preview` is fully wired (allow-list, DB constraint, pricing, region routing, thinking config) but is **not** the default production model — live concurrent-load testing found it returns 429s above ~10 concurrent requests on the current GCP project quota. `gemini-3.1-flash-lite` did not show that ceiling across sequential (15 calls) and concurrent (5/10 concurrent) live testing and is the recommended model of the two for production traffic today.

## Test plan

- [x] Full test suite green (`pytest tests/ -q`)
- [x] `ruff check .` clean
- [x] Live-verified against the real Vertex AI API (not just mocked): region routing, `thinking_config` behavior, and the empty-response failure mode this PR fixes
- [x] Live streaming latency + stability benchmarks for `gemini-3.1-flash-lite` (15 sequential + up to 10 concurrent, zero errors)
- [x] Live concurrent-load test for `gemini-3-flash-preview` isolating its 429 ceiling (~10 concurrent)

### DIFF
--- a/alembic/versions/20260816_widen_agent_llm_model_check_gemini3.py
+++ b/alembic/versions/20260816_widen_agent_llm_model_check_gemini3.py
@@ -1,0 +1,140 @@
+"""widen ck_agent_llm_model to include gemini-3 family
+
+Revision ID: 20260816_widen_llm_check_g3
+Revises: 20260815_widen_llm_check
+Create Date: 2026-08-16
+
+Purpose
+-------
+``app/core/llm_models.py`` (the Python-level allow-list used to validate
+``Agent.llm_model`` at the API layer) now includes two new Google Gemini
+models: ``gemini-3-flash-preview`` (Gemini 3 Flash Preview, Vertex AI
+preview) and ``gemini-3.1-flash-lite`` (Gemini 3.1 Flash-Lite, GA as of
+May 2026). The DB-level ``ck_agent_llm_model`` CHECK constraint — most
+recently widened by ``20260815_widen_llm_check`` to a 19-value snapshot —
+was never widened to match, so inserting/updating an agent with either of
+these new model IDs would violate the CHECK even though the application
+layer accepts them.
+
+This migration is purely additive: it drops and recreates
+``ck_agent_llm_model`` with the union of the prior 19-value snapshot plus
+the 2 new gemini-3 IDs (21 total). No values are removed or renamed, so no
+existing row can violate the widened constraint — safe to apply without a
+backfill on a live multi-tenant table. The drop+recreate briefly requires a
+table-level lock while Postgres validates the new CHECK against existing
+rows, but on a boolean/string membership check on `agent` (not a huge
+table) this is expected to be fast; the definition change itself is
+otherwise the same additive-CHECK-widening pattern used by
+``20260815_widen_llm_check`` and, before it, ``20260602_schema_v2_completion``
+for ``ck_agent_status_v2`` / ``ck_callflow_direction``.
+
+Snapshot strategy
+------------------
+Per the same rationale as ``20260815_widen_llm_check``, the allow-list used
+to build the CHECK SQL is a self-contained snapshot tuple in *this* file,
+not an import from ``app.core.llm_models`` — so future edits to that module
+(renames, removals) do not retroactively change what this historical
+migration replays during ``alembic upgrade``.
+"""
+from __future__ import annotations
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260816_widen_llm_check_g3"
+down_revision: Union[str, Sequence[str], None] = "20260815_widen_llm_check"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+# Union of the 20260815_widen_llm_check 19-value snapshot plus the 2 new
+# Gemini 3 family IDs added to app.core.llm_models.ALLOWED_LLM_MODELS.
+# Self-contained — do not import app.core.llm_models (future renames there
+# must not retroactively change this migration's replay behavior).
+_ALLOWED_LLM_MODELS_AT_REVISION: tuple[str, ...] = (
+    # --- prior 19-value snapshot (20260815_widen_llm_check), unchanged ---
+    "gpt-4o-mini",
+    "gpt-4o",
+    "gpt-4.1",
+    "gpt-4.1-mini",
+    "gpt-4-turbo",
+    "gemini-2.5-flash",
+    "gemini-2.0-flash-001",
+    "gemini-2.0-flash",
+    "gemini-1.5-pro",
+    "gemini-1.5-flash",
+    "claude-3-5-sonnet",
+    "claude-3-haiku",
+    "llama-3.1-70b-versatile",
+    "llama-3.1-8b-instant",
+    "gpt-5",
+    "gpt-5-mini",
+    "gpt-5.1",
+    "gpt-5.2",
+    "gpt-5.4",
+    # --- new: Google Gemini 3 family ---
+    "gemini-3-flash-preview",
+    "gemini-3.1-flash-lite",
+)
+
+# The exact 19-value snapshot from 20260815_widen_llm_check, needed to
+# restore the narrower constraint on downgrade.
+_PRIOR_ALLOWED_LLM_MODELS: tuple[str, ...] = (
+    "gpt-4o-mini",
+    "gpt-4o",
+    "gpt-4.1",
+    "gpt-4.1-mini",
+    "gpt-4-turbo",
+    "gemini-2.5-flash",
+    "gemini-2.0-flash-001",
+    "gemini-2.0-flash",
+    "gemini-1.5-pro",
+    "gemini-1.5-flash",
+    "claude-3-5-sonnet",
+    "claude-3-haiku",
+    "llama-3.1-70b-versatile",
+    "llama-3.1-8b-instant",
+    "gpt-5",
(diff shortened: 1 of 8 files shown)


app/services/vertex_gemini_service.py
```diff
@@ -18,27 +22,82 @@
...
+# first spoken word). MUST be "minimal", not "low": thinking tokens are
+# deducted from max_output_tokens on this model, and this service's callers
+# use small voice-turn budgets (max_tokens=100 for stream_text/generate_text,
+# 200 for generate_with_tools). Verified against the live API: at "low",
+# max_tokens=60 produced 57 thinking tokens and ZERO visible text
+# (finish_reason=MAX_TOKENS, empty response — the caller would go silent
+# mid-call); even max_tokens=100 left only 4 tokens for actual output. At
+# "minimal", thinking is negligible (thoughts_token_count=None) and the full
+# answer is reliably produced well within a 60-100 token budget. Do not
+# change this to "low"/"medium"/"high" without also verifying against the
```
codewithwaleed: 🔴 **Blocker: sync `client.models.generate_content()` blocks the event loop**

`stream_text` and `generate_with_tools` both use `await client.aio.models.*`. This sync call stalls the event loop for the full LLM round-trip on every non-streaming Gemini call, starving concurrent voice sessions.

```python
# Fix: match stream_text/generate_with_tools pattern
async def generate_text(self, ...):
    ...
    response = await client.aio.models.generate_content(
        model=model_name,
        contents=contents,
        config=generation_config,
    )
    response_text = (response.text or "").strip()
```

app/services/pricing_service.py
```diff
@@ -52,6 +52,9 @@ def __init__(self):
             "gemini-2.5-flash-lite": (0.05, 0.20),
             "gemini-2.0-flash": (0.05, 0.20),
             "gemini-2.0-flash-lite": (0.04, 0.15),
+            # Placeholder — same rate as gemini-2.5-flash pending official preview pricing.
+            "gemini-3-flash-preview": (0.15, 1.25),
```
codewithwaleed: 🟠 **Major: Placeholder rates don't match the stated reference**

Comment says "same rate as gemini-2.5-flash" but `(0.15, 1.25)` / `(0.25, 1.50)` are 2–4× higher than `gemini-2.5-flash`'s rate in this table. Tenants will be over-billed.

```python
# Use the actual gemini-2.5-flash rates until official pricing is announced:
"gemini-3-flash-preview": (0.075, 0.30),   # TODO: update when GA pricing confirmed
"gemini-3.1-flash-lite": (0.075, 0.30),
```

alembic/versions/20260816_widen_agent_llm_model_check_gemini3.py
```diff
@@ -0,0 +1,140 @@
...
+    "gpt-5",
+    "gpt-5-mini",
+    "gpt-5.1",
+    "gpt-5.2",
+    "gpt-5.4",
+)
+
+
+def _llm_check_sql(models: tuple[str, ...]) -> str:
+    """Build ck_agent_llm_model CHECK SQL from a given model snapshot."""
```
codewithwaleed: 🟠 **Major: SQL string interpolation without format validation**

`_llm_check_sql` interpolates model names directly via `f"'{m}'"`. All current names are safe, but a future name with a single quote generates broken or injectable SQL that Alembic executes against the DB.

```python
import re

def _llm_check_sql(models: tuple[str, ...]) -> str:
    for m in models:
        if not re.fullmatch(r"[a-zA-Z0-9._-]+", m):
            raise ValueError(f"Unsafe model name for SQL: {m!r}")
    return "llm_model IS NULL OR llm_model IN (%s)" % ", ".join(f"'{m}'" for m in models)
```

app/services/vertex_gemini_service.py
```diff
@@ -18,27 +22,82 @@
```
codewithwaleed: 🟡 **Minor: `str(finish)` comparison is SDK-repr-dependent**

`str(finish) not in ("STOP", "MAX_TOKENS", "1", "2")` — if the google-genai SDK changes how it serializes `FinishReason` (e.g. `"FinishReason.STOP"` instead of `"STOP"`), every clean completion silently becomes a content-filter error mid-call.

```python
_CLEAN_FINISH_REASONS = {
    "STOP", "MAX_TOKENS", "1", "2",
    "FINISH_REASON_STOP", "FINISH_REASON_MAX_TOKENS",
}
if finish and str(finish).upper() not in _CLEAN_FINISH_REASONS:
    return VertexLlmError(...)
```

codewithwaleed (changes requested):
## PR #209 Review — Migrate Vertex Gemini service to google-genai SDK; add Gemini 3 model support

### Summary
This PR migrates `vertex_gemini_service.py` from the deprecated `vertexai.generative_models` SDK to `google-genai` (`google.genai`), adds Gemini 3 family model support with correct thinking-level pinning, per-location client caching for the `global`-only Gemini 3 endpoint, and keeps the migration's snapshot strategy consistent. The test rewrite is thorough — `_real_google_genai()` for selectively loading the real SDK is a clean pattern, and the `TestThinkingConfigRouting` tests correctly use the real SDK rather than a MagicMock to assert on actual field values.

Issues require attention before merge.

---

### 🔴 Blocker

**`generate_text` uses sync `client.models.generate_content()` — blocks the event loop**

`stream_text` and `generate_with_tools` correctly use `await client.aio.models.*`. But `generate_text` still calls the sync `client.models.generate_content()` — meaning every non-streaming LLM call on Gemini blocks the event loop for the full round-trip (~500ms–2s), starving all concurrent voice sessions on that worker.

```python
# Fix: change to async and use client.aio
async def generate_text(self, ...):
    ...
    response = await client.aio.models.generate_content(
        model=model_name,
        contents=contents,
        config=generation_config,
    )
```

---

### 🟠 Major Issues

**Placeholder pricing for Gemini 3 doesn't match the stated reference model**

The comment says "same rate as gemini-2.5-flash" but the actual rates `(0.15, 1.25)` and `(0.25, 1.50)` are 2–4× higher than `gemini-2.5-flash`'s rate in this same table. Tenants using these models will be over-billed until corrected. Either use the actual `gemini-2.5-flash` rates or add a tracking reference:
```python
# TODO: update when official pricing announced — placeholder uses gemini-2.5-flash rates
"gemini-3-flash-preview": (0.075, 0.30),
"gemini-3.1-flash-lite": (0.075, 0.30),
```

**`_llm_check_sql` interpolates model names directly into SQL**

All current model names are safe, but this pattern is fragile for future additions. A model name with a special character generates broken or injectable SQL executed directly by Alembic. Add validation:
```python
import re
def _llm_check_sql(models):
    for m in models:
        if not re.fullmatch(r"[a-zA-Z0-9._-]+", m):
            raise ValueError(f"Unsafe model name: {m!r}")
    return "llm_model IS NULL OR llm_model IN (%s)" % ", ".join(f"'{m}'" for m in models)
```

---

### 🟡 Minor Issues

**`finish_reason` string comparison may break if SDK enum repr changes**

`str(finish) not in ("STOP", "MAX_TOKENS", "1", "2")` assumes specific string forms. Some SDK versions use `"FinishReason.STOP"`. A mismatch silently turns every clean completion into a content-filter error mid-call.
```python
_CLEAN_FINISH_REASONS = {"STOP", "MAX_TOKENS", "1", "2", "FINISH_REASON_STOP", "FINISH_REASON_MAX_TOKENS"}
if finish and str(finish).upper() not in _CLEAN_FINISH_REASONS:
```

**`_GLOBAL_ONLY_MODELS` set maintenance is asymmetric** — models need to be added when new Gemini 3 submodels ship on global-only, and removed when regional availability arrives. The comment mentions one direction but not both. A short note would help future maintainers.

---

### Positives
- `_real_google_genai()` context manager pattern is clean and properly restores `sys.modules` in `finally`. ✓
- `role="user"` for function-response content (replacing `role="function"`) is correctly justified by the `google.genai` AFC loop convention. ✓
- `Part.from_text(text=text)` keyword-only fix is correct and backed by a regression test. ✓
- Thinking-level explicitly pinned (`"MINIMAL"`) for Gemini 3 family — correct and well-justified given the shared `max_output_tokens` budget. ✓

---

### Recommendation: **Request Changes** — fix the sync/blocking `generate_text` call before merge; confirm or correct the pricing rates.

PR #210, opened 2026-08-17
Gemini Live native-audio speech-to-speech models

Aliict:
## Summary

Integrates Gemini Live native-audio speech-to-speech models as a new pipeline path alongside the existing STT→LLM→TTS pipeline, wired into both call transports (Twilio bidirectional stream and browser/LiveKit).

- **`28445d8`** — Core integration: `GeminiLiveSession` wrapper (`app/services/gemini_live_service.py`), audio format bridging, `GEMINI_LIVE_MODELS` routing, wired into `VoiceOrchestrator` for both transports.
- **`dff58ee`** — Phase 2: Calendly tool-calling, mid-call RAG context refresh, voice picker.
- **`a39d10b`** — Fix: `_receive_loop()` only processed the first Gemini conversation turn — `session.receive()` covers a single turn by SDK design and must be re-invoked per turn, or the receive loop dies silently after turn 1, leaving calls fully silent until a keepalive timeout tears down the connection. Confirmed working on a real call after this fix.
- **`564b9b8`** — Fix: input/output transcript capture was broken — Gemini's `finished` flag on transcription messages is not reliably set (confirmed via an open upstream bug, googleapis/js-genai#1429) and even when set, `text` is often just an incremental fragment. Replaced with fragment-buffering flushed on reliable signals (turn_complete, next-turn's first output fragment, barge-in, shutdown) instead of trusting `finished`.
- **`b2ede57`** — Added regression coverage that was missing for the receive-loop multi-turn fix (a retrospective review found the existing suite never exercised a second turn), plus a genuine mid-`receive()` cancellation test.

## Test plan

- [x] `pytest tests/voice/test_gemini_live_service.py tests/voice/test_gemini_live_twilio_fork.py tests/voice/test_gemini_live_browser_fork.py -q` — 117 passed
- [x] Verified live on real phone calls (both bugs above were found and fixed via real-call testing, not just unit tests)
- [x] Retrospective `code-reviewer` pass on both fix commits, findings addressed inline

### DIFF
--- a/app/services/gemini_live_service.py
+++ b/app/services/gemini_live_service.py
@@ -241,12 +241,30 @@ class GeminiLiveSession:
         """Send one chunk of raw PCM16/16kHz/mono caller audio. Caller must
         have already converted to this format (e.g. via
         ``gemini_live_audio_bridge.mulaw8k_to_pcm16_16k``) — this module has
-        zero telephony-format knowledge."""
+        zero telephony-format knowledge.
+
+        Safe no-op once the session is closed/closing — the audio-forwarding
+        background task (``LiveKitAudioSubscriber`` / Twilio's media-stream
+        loop) is stopped independently of this session's teardown, so a few
+        straggler chunks arriving after ``close()`` has already fired are
+        expected, not exceptional. Without this guard those stragglers hit
+        the underlying websocket mid-close-handshake and raise
+        ``ConnectionClosedError``/``TimeoutError``, which the caller then
+        logs at ERROR level for what is actually a benign shutdown race."""
+        if not pcm16_16k_bytes or self._closed:
+            return
         from google.genai import types
 
-        await self._session.send_realtime_input(
-            audio=types.Blob(data=pcm16_16k_bytes, mime_type="audio/pcm;rate=16000")
-        )
+        try:
+            await self._session.send_realtime_input(
+                audio=types.Blob(data=pcm16_16k_bytes, mime_type="audio/pcm;rate=16000")
+            )
+        except Exception:
+            if self._closed:
+                # Lost the race: close() ran concurrently with this send.
+                # Benign — swallow rather than surface as an error.
+                return
+            raise
 
     async def send_text(self, text: str, turn_complete: bool = True) -> None:
         from google.genai import types
--- a/app/services/gemini_live_service.py
+++ b/app/services/gemini_live_service.py
@@ -305,12 +305,16 @@ class GeminiLiveSession:
                     except Exception as exc:
                         await self._report_error(exc)
                 if turn_msg_count == 0:
-                    # receive() returned with zero messages instead of raising —
-                    # treat as a dead/closed connection rather than busy-loop
-                    # re-invoking it forever.
+                    # Defensive guard, not currently reachable against the pinned
+                    # google-genai SDK: a dead/closed connection makes receive()
+                    # raise (handled by the except below), not return an empty
+                    # generator. This only exists in case a future SDK version
+                    # ever yields zero messages instead of raising — without it,
+                    # re-invoking receive() forever in that case would busy-spin
+                    # the event loop rather than ending the call gracefully.
                     logger.warning(
                         "[GeminiLiveSession] receive() yielded no messages model=%s — "
-                        "ending receive loop (connection likely closed)",
+                        "ending receive loop",
                         self.model_name,
                     )
                     break
--- a/app/services/gemini_live_service.py
+++ b/app/services/gemini_live_service.py
@@ -1,0 +1,301 @@
+"""
+Gemini Live (speech-to-speech native-audio) session wrapper.
+
+Mirrors ``app/services/vertex_gemini_service.py``'s module shape but wraps a
+*stateful* Live session instead of one-shot request/response calls. This
+module has ZERO knowledge of Twilio/LiveKit/MULAW/telephony — it only speaks
+raw PCM16 bytes in (16kHz, caller -> Gemini) and out (24kHz, Gemini ->
+caller), exactly like ``vertex_gemini_service.py`` has zero knowledge of any
+TTS/STT provider. Byte-level format conversion to/from telephony formats
+lives in ``app/voice/gemini_live_audio_bridge.py``; wiring this session into
+an actual call transport is out of scope for this module (see
+``app/core/llm_models.py``'s ``GEMINI_LIVE_MODELS``/
+``is_gemini_live_native_audio_model`` for the routing predicate used by
+transport handlers).
+
+Auth is per-model (see ``GEMINI_LIVE_MODELS``): two models use Vertex AI/ADC
+(reusing ``vertex_gemini_service._ensure_vertex_client`` verbatim — same
+client type, same location-keyed cache, no duplication), one uses the
+Gemini Developer API with ``settings.GEMINI_API_KEY`` as a deliberate, scoped
+exception (see that constant's docstring for why).
+"""
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import threading
+from typing import Any, Awaitable, Callable, Sequence
+
+from app.core.config import settings
+from app.core.llm_models import GEMINI_LIVE_MODELS
+from app.core.logger import logger
+from app.services.vertex_gemini_service import (
+    VertexLlmError,
+    VertexLlmErrorType,
+    _classify_vertex_error,
+    _ensure_vertex_client,
+)
+
+DEFAULT_VOICE_NAME = "Puck"
+
+_api_key_client_lock = threading.Lock()
+_api_key_client: Any = None
+
+
+def _ensure_api_key_client():
+    """
+    Lazily build (once, process-wide) and return a shared google.genai
+    Client configured for the Gemini Developer API (API-key auth) — used
+    only by ``gemini-3.1-flash-live-preview`` (see ``GEMINI_LIVE_MODELS``).
+    Unlike ``_ensure_vertex_client``, this is NOT location-keyed: the
+    Developer API is not project/location-scoped.
+    """
+    global _api_key_client
+    if _api_key_client is not None:
+        return _api_key_client
+    with _api_key_client_lock:
+        if _api_key_client is not None:
+            return _api_key_client
+        from google import genai
+
+        api_key = settings.GEMINI_API_KEY
+        if not api_key:
+            raise RuntimeError(
+                "GEMINI_API_KEY is required for Gemini Live Developer-API models "
+                "(e.g. gemini-3.1-flash-live-preview)."
+            )
+        _api_key_client = genai.Client(api_key=api_key)
+        return _api_key_client
+
+
+class GeminiLiveSession:
+    """
+    Transport-agnostic wrapper around one ``google.genai`` Live API session.
+
+    Usage (by a future transport-handler wiring, not part of this task):
+      session = GeminiLiveSession(agent.llm_model)
+      await session.start(system_instruction=..., on_audio_chunk=..., on_interrupted=...)
+      await session.send_audio(pcm16_16k_bytes)   # per caller-audio chunk
+      ...
+      await session.close()
+    """
+
+    def __init__(self, model_name: str) -> None:
+        config = GEMINI_LIVE_MODELS.get(model_name)
+        if config is None:
+            raise ValueError(
+                f"'{model_name}' is not a Gemini Live native-audio model. "
+                f"Valid models: {sorted(GEMINI_LIVE_MODELS)}"
+            )
+        self.model_name = model_name
+        self._auth_mode: str = config["auth"]
+        self._location: str | None = config["location"]
+
+        self._client: Any = None
+        self._connect_cm: Any = None
+        self._session: Any = None
+        self._receive_task: asyncio.Task | None = None
+        self._closed = False
+
+        # Registered by start(); consumed by _receive_loop.
+        self.on_audio_chunk: Callable[[bytes], Awaitable[None]] | None = None
+        self.on_interrupted: Callable[[], Awaitable[None]] | None = None
+        self.on_input_transcript: Callable[[str, bool], Awaitable[None]] | None = None
+        self.on_output_transcript: Callable[[str, bool], Awaitable[None]] | None = None
+        self.on_tool_call: Callable[[Sequence[Any]], Awaitable[None]] | None = None
+        self.on_error: Callable[[VertexLlmError], Awaitable[None]] | None = None
+
+    def _build_client(self):
+        if self._auth_mode == "vertex":
+            return _ensure_vertex_client(self._location)
+        if self._auth_mode == "api_key":
+            return _ensure_api_key_client()
+        raise VertexLlmError(
+            f"Unknown Gemini Live auth mode '{self._auth_mode}' for model {self.model_name}",
+            VertexLlmErrorType.UNKNOWN,
+        )
+
+    async def start(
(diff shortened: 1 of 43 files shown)


app/services/gemini_live_service.py
```diff
@@ -0,0 +1,402 @@
...
+        live_config = types.LiveConnectConfig(**config_kwargs)
+
+        try:
+            self._connect_cm = self._client.aio.live.connect(
+                model=self.model_name, config=live_config
+            )
+            self._session = await self._connect_cm.__aenter__()
+        except Exception as exc:
+            self._connect_cm = None
+            raise _classify_vertex_error(exc) from exc
```
codewithwaleed: 🔴 **Blocker: `self._closed = True` is set before teardown completes — `send_audio`'s exception guard fires too early**

`close()` sets `self._closed = True` first, then awaits `receive_task.cancel()`. During that window, `send_audio`'s `except Exception: if self._closed: return` guard will silently swallow ANY exception — not just the expected teardown-race `ConnectionClosedError`.

If the session has a bug (bad `Blob` field, etc.) that raises during teardown, the error is silently dropped.

Fix — check specifically for connection-closed exceptions rather than all `Exception`:

```python
except Exception as exc:
    if self._closed and type(exc).__name__ in (
        "ConnectionClosedError", "ConnectionClosedOK",
        "TimeoutError", "CancelledError",
    ):
        return  # benign shutdown race
    raise
```

app/voice/voice_orchestrator.py
```diff
@@ -456,6 +566,37 @@ async def shutdown(self) -> None:
...
+        # exactly the "call ending mid-buffer" case this flush exists to cover.
+        try:
+            if self._gemini_live_session is not None:
+                self._gemini_live_cancel.set()
+                await self._gemini_live_session.close()
+                self._gemini_live_session = None
+                await self._flush_gemini_live_input_buffer()
+                await self._flush_gemini_live_output_buffer()
+        except Exception as exc:
+            logger.debug("[VoiceOrchestrator] Gemini Live session close failed: %s", exc)
```
codewithwaleed: 🔴 **Blocker: `agent.llm_model = _GEMINI_LIVE_FALLBACK_TEXT_MODEL` mutates the SQLAlchemy ORM object in-place — may be flushed to the DB at call end**

If the `db` session is still open and a `commit()`/`flush()` happens later (e.g. in `CallSessionService.update_call_session_status()`), SQLAlchemy will detect the dirty `agent` object and persist the fallback text model name to the `agent` row — permanently changing the agent's configured LLM model in the database for ALL future calls.

Fix — expunge the agent from the session before mutating, or store the override on `VoiceOrchestrator` instead:

```python
# Option A: expunge before mutating so SQLAlchemy won't flush it
if hasattr(h, "db") and h.db:
    try: h.db.expunge(agent)
    except Exception: pass
agent.llm_model = _GEMINI_LIVE_FALLBACK_TEXT_MODEL

# Option B: store override on self, not the ORM object
self._effective_llm_model = _GEMINI_LIVE_FALLBACK_TEXT_MODEL
```

app/services/gemini_live_service.py
```diff
@@ -0,0 +1,402 @@
...
+        try:
+            # session.receive() is an async-generator that, per google-genai's
+            # live.py implementation, yields only until the first
+            # server_content.turn_complete=True message and then returns
+            # (StopAsyncIteration) — it covers a single model turn, not the
+            # whole call. It must be re-invoked for every subsequent turn,
+            # hence the outer while loop below.
+            while True:
+                turn_msg_count = 0
+                async for message in self._session.receive():
```
codewithwaleed: 🟠 **Major: `turn_msg_count == 0` silently exits the receive loop without firing `on_error`**

If a future SDK returns an empty generator (the guard's rationale), the call goes silent with no error callback — the caller never knows the receive loop ended.

Add an error report before breaking:

```python
if turn_msg_count == 0:
    logger.warning("[GeminiLiveSession] receive() yielded no messages model=%s — ending receive loop", self.model_name)
    await self._report_error(
        VertexLlmError("receive() returned empty turn", VertexLlmErrorType.UNKNOWN)
    )
    break
```

app/voice/voice_orchestrator.py
```diff
@@ -516,6 +657,882 @@ async def _run_final() -> None:
...
+    # ── Gemini Live (native-audio speech-to-speech) ────────────────────────────
+
+    async def _feed_gemini_live_audio(self, h, audio_data: bytes) -> None:
+        """
+        Route one MULAW/8kHz caller-audio frame to the (lazily created)
+        GeminiLiveSession for this call, converting to PCM16/16kHz first.
+        Called instead of the SttPipeline feed for native-audio agents.
+        """
+        if self._gemini_live_setup_failed:
+            # Pre-session failure already handled (see _fallback_to_legacy_pipeline) —
```
codewithwaleed: 🟠 **Major: `asyncio.create_task(h._full_shutdown())` has no reference kept — fire-and-forget task**

If the event loop tears down before `_full_shutdown()` completes, this task is silently cancelled. Keep a reference:

```python
# store so it isn't garbage-collected / silently cancelled
self._gemini_live_session_task = asyncio.create_task(h._full_shutdown())
```

app/services/gemini_live_service.py
```diff
@@ -0,0 +1,402 @@
...
+Mirrors ``app/services/vertex_gemini_service.py``'s module shape but wraps a
+*stateful* Live session instead of one-shot request/response calls. This
+module has ZERO knowledge of Twilio/LiveKit/MULAW/telephony — it only speaks
+raw PCM16 bytes in (16kHz, caller -> Gemini) and out (24kHz, Gemini ->
+caller), exactly like ``vertex_gemini_service.py`` has zero knowledge of any
+TTS/STT provider. Byte-level format conversion to/from telephony formats
+lives in ``app/voice/gemini_live_audio_bridge.py``; wiring this session into
+an actual call transport is out of scope for this module (see
+``app/core/llm_models.py``'s ``GEMINI_LIVE_MODELS``/
+``is_gemini_live_native_audio_model`` for the routing predicate used by
```
codewithwaleed: 🟡 **Minor: `resolve_voice_name` linear scan — use a pre-built case-folded dict**

```python
_VOICE_NAME_LOWER: dict[str, str] = {v.lower(): v for v in VALID_VOICE_NAMES}

def resolve_voice_name(requested: str | None) -> str:
    if not requested:
        return DEFAULT_VOICE_NAME
    resolved = _VOICE_NAME_LOWER.get(requested.strip().lower())
    if resolved:
        return resolved
    logger.info("[GeminiLive] voice '%s' not valid — falling back to %s", requested, DEFAULT_VOICE_NAME)
    return DEFAULT_VOICE_NAME
```
Same pattern applies to `openai_realtime_service.resolve_voice_name`.

app/services/gemini_live_service.py
```diff
@@ -0,0 +1,402 @@
...
+                )
+
+            output_transcript = getattr(server_content, "output_transcription", None)
+            if output_transcript is not None and self.on_output_transcript is not None:
+                await self.on_output_transcript(
+                    getattr(output_transcript, "text", "") or "",
+                    bool(getattr(output_transcript, "finished", False)),
+                )
+
+            # Wire-protocol turn boundary: per-fragment ``finished`` flags on
```
codewithwaleed: 🔵 **Nit: `go_away` handler does nothing ("phase 1") — confirm there's a follow-up plan**

`go_away` means the server is about to close — doing nothing means the next `recv()` will fail unexpectedly rather than the session proactively closing. If phase 2 will add a clean shutdown here, a TODO comment with a ticket reference would help:

```python
if getattr(message, "go_away", None):
    logger.warning("[GeminiLiveSession] go_away received model=%s — TODO: proactive close (phase 2)", self.model_name)
    # Phase 2: call self.close() and signal the transport to reconnect or end gracefully
```

codewithwaleed (changes requested):
## PR #210 Review — Gemini Live native-audio speech-to-speech models

### Summary
This is a substantial PR that adds two new real-time speech-to-speech providers — Gemini Live (Google) and OpenAI Realtime — as drop-in alternatives to the existing STT/LLM/TTS pipeline. The architecture is well-designed: both providers share a common callback contract (`on_audio_chunk`, `on_interrupted`, `on_turn_complete`, `on_tool_call`, `on_error`) so `VoiceOrchestrator` can route either provider into the same transport handlers. The lazy-session-on-first-audio-chunk pattern with fallback to the legacy pipeline on pre-session failure is particularly well-executed. The PR is extensively commented and the test coverage for both new services is thorough.

---

### 🔴 Blockers

**1. `GeminiLiveSession.send_audio` silently swallows non-close exceptions when `self._closed` is True — but `self._closed` is set at the *start* of `close()`, before teardown completes**

```python
# close() sets self._closed = True FIRST, then awaits receive_task.cancel()
self._closed = True
if self._receive_task is not None:
    self._receive_task.cancel()
    with contextlib.suppress(...): await self._receive_task  # ← teardown still in progress
```

Meanwhile, `send_audio` checks `self._closed` as a guard:

```python
except Exception:
    if self._closed:
        return  # swallowed
    raise
```

Any exception during the still-in-progress teardown window is swallowed — not just the expected `ConnectionClosedError`. If the session has a bug (e.g. a bad `types.Blob` field) that raises before the close actually completes, the error will be silently dropped with no log. The guard is correct for `ConnectionClosedError` and `TimeoutError` but too broad for other exceptions.

Fix — catch the specific expected teardown-race exceptions rather than all `Exception`:

```python
except (ConnectionError, asyncio.TimeoutError, Exception) as exc:
    if self._closed and ("closed" in str(exc).lower() or "timeout" in str(exc).lower()):
        return  # lost the close race — benign
    raise
```

**2. `_fallback_to_legacy_pipeline` mutates `agent.llm_model` directly on the ORM instance — this is an in-memory mutation that bypasses SQLAlchemy's dirty-tracking and could be flushed back to the DB**

```python
agent.llm_model = _GEMINI_LIVE_FALLBACK_TEXT_MODEL  # mutates the loaded ORM object
```

If the `db` session is still open anywhere in the call path and a `db.commit()` or `db.flush()` happens after this mutation (e.g. inside `CallSessionService.update_call_session_status()` at call end), SQLAlchemy will detect the dirty `agent` object and persist the fallback model name to the `agent` row — permanently changing the agent's configured model in the database to the fallback text model for all future calls.

Fix — use `db.expunge(agent)` or only mutate a detached copy:

```python
# Option A: expunge the object from the session so SQLAlchemy won't flush it
if hasattr(h, "db") and h.db:
    try:
        h.db.expunge(agent)
    except Exception:
        pass
agent.llm_model = _GEMINI_LIVE_FALLBACK_TEXT_MODEL

# Option B: store the fallback model on VoiceOrchestrator only, not the ORM object
self._effective_llm_model = _GEMINI_LIVE_FALLBACK_TEXT_MODEL
```

---

### 🟠 Major Issues

**3. `_receive_loop` in `GeminiLiveSession` breaks the outer `while True` loop when `turn_msg_count == 0` — but this is called "defensive" and "not currently reachable"**

The comment correctly identifies that the current google-genai SDK always raises on a dead connection rather than returning an empty generator. But if this guard fires on a live SDK version that does return empty (e.g. a fast reconnect scenario), the whole receive loop silently exits, leaving `on_audio_chunk` / `on_turn_complete` never called again for the rest of the call. The caller gets silence with no error.

The `while True` loop should `await self._report_error(...)` before breaking in this case, not just log and break:

```python
if turn_msg_count == 0:
    logger.warning("[GeminiLiveSession] receive() yielded no messages ...")
    await self._report_error(VertexLlmError("receive() returned empty turn", VertexLlmErrorType.UNKNOWN))
    break
```

**4. `OpenAIRealtimeSession` (inferred from `openai_realtime_service.py`) uses `base64.b64decode` on incoming audio chunks — if the Realtime API sends audio as binary frames (not base64-encoded JSON), this will silently produce garbled audio**

From the module docstring and constants (`TWILIO_AUDIO_FORMAT = "audio/pcmu"`), it's clear the API is expected to send binary audio. But the `_receive_loop` for OpenAI Realtime should be confirmed to handle both `bytes` frames and `str`/base64 JSON frames — the same distinction that caused the `HumeTtsService` binary-frame crash (PR #212 issue, now documented). Please add a comment confirming which format the `openai` SDK's `AsyncRealtimeConnection.recv()` returns, and whether the `base64` import is actually used.

---

### 🟡 Minor Issues

**5. `resolve_voice_name` in both `gemini_live_service.py` and `openai_realtime_service.py` uses a linear scan through `VALID_VOICE_NAMES` for a case-insensitive lookup — O(n) on every call**

For a frozenset of 10-30 names this is negligible, but a case-folded lookup dict would be cleaner and eliminate the risk of a future maintainer removing the linear scan "optimization":

```python
_VOICE_NAME_LOWER: dict[str, str] = {v.lower(): v for v in VALID_VOICE_NAMES}

def resolve_voice_name(requested: str | None) -> str:
    if not requested:
        return DEFAULT_VOICE_NAME
    resolved = _VOICE_NAME_LOWER.get(requested.strip().lower())
    if resolved:
        return resolved
    logger.info("[GeminiLive] voice '%s' not valid — falling back to %s", requested, DEFAULT_VOICE_NAME)
    return DEFAULT_VOICE_NAME
```

**6. `VoiceOrchestrator._fallback_to_legacy_pipeline` calls `asyncio.create_task(h._full_shutdown())` in the "no agent loaded" branch — this is a fire-and-forget task with no reference kept**

If the event loop is torn down before `_full_shutdown()` completes, the task is silently cancelled with no log. Assign the task to a tracked attribute or use `asyncio.shield`:

```python
task = asyncio.create_task(h._full_shutdown())
self._gemini_live_session_task = task  # keep a reference
```

**7. The `go_away` handler in `GeminiLiveSession._handle_message` logs and does nothing ("no action taken (phase 1)")**

A `go_away` message means the server is about to close the connection — typically a deployment/maintenance event. Doing nothing means the session will silently fail on the next `recv()` call rather than proactively closing and triggering the error path cleanly. A `phase 1` comment implies this is intentional deferral, which is fine. Just confirm there's a `phase 2` plan to handle it.

---

### Missing Test Coverage

- **`_fallback_to_legacy_pipeline` ORM mutation test**: a test that creates a mock ORM `Agent`, runs `_fallback_to_legacy_pipeline`, then checks whether the original model name is still in the DB session's dirty set — this would catch the unintentional-flush issue (#2).
- **`GeminiLiveSession.close()` then `send_audio()` race**: test that `send_audio()` called concurrently with `close()` doesn't permanently swallow non-teardown exceptions.
- **`resolve_voice_name` with an ElevenLabs voice ID**: both services need a test that an ElevenLabs-style voice ID (e.g. `"21m00Tcm4TlvDq8ikWAM"`) correctly falls back to the default and logs.

---

### Positives
- The `_receive_loop` outer `while True` and the `turn_msg_count` guard is well-motivated — it reflects real experience with a Gemini Live "silent after turn 1" production bug. ✓
- The deliberate divergences between `GeminiLiveSession` and `OpenAIRealtimeSession` are thoroughly documented (per-turn generator re-invocation vs. flat receive loop; transcript buffering vs. per-event `completed`). ✓
- Pre-session fallback to the legacy pipeline is the correct design — a failed Gemini Live session should not drop the call entirely. ✓
- `VALID_VOICE_NAMES` with case-insensitive lookup prevents bad voice IDs from 400-ing the session. ✓

---

### Recommendation: **Request Changes** — fix the ORM mutation (#2) before merge; the `send_audio` exception guard (#1) is a lower-risk correctness issue but worth addressing.

PR #212, opened 2026-08-18
feat(voice): add Hume AI as a real-time streaming TTS provider

Aliict:
## Summary
- Adds Hume AI as a new production-ready, real-time streaming TTS provider alongside ElevenLabs/Google/Rime, following the existing adapter architecture (`HumeTTSAdapter`, `hume_tts_service.py`, capability table, config/secret plumbing, catalog registration).
- Fixes a production incident found after deploying/testing against Hume's live WebSocket endpoint: Hume sends some audio as raw **binary** WS frames rather than the documented JSON+base64 envelope, which crashed the receive loop with `UnicodeDecodeError`. Binary frames are now decoded directly.

## Changes
- `app/services/hume_tts_service.py` — WebSocket client to Hume's `stream-input` endpoint, one connection per TTS chunk, PCM→mulaw 8kHz incremental downsampling, timeout/cancellation/error handling, and the binary-frame fix.
- `app/utils/tts_adapter.py` — `HumeTTSAdapter` + dedicated sync event-loop bridge, registered in `get_tts_adapter()`.
- `app/utils/audio_utils.py` — generic `PCMStreamDownsampler(src_rate_hz, dst_rate_hz)`.
- `app/voice/tts_provider_capabilities.py`, `app/schemas/agent.py`, `app/core/agent_runtime.py`, `app/core/config.py`, `app/core/secret_manager.py`, `.env.example`, `app/main.py`, `app/services/tts_catalog_service.py`, `app/voice/tts_stream_mixin.py`, `app/voice/livekit_browser_call_handler.py` — provider registration, default-voice fallback (matching Rime's precedent), catalog seeding, startup fail-fast check.
- Tests: `tests/voice/test_hume_tts_and_barge_in.py`, `tests/core/test_secret_manager_hume.py`, plus additions to `tests/utils/test_audio_utils.py` / `tests/voice/test_tts_provider_capabilities.py`.

## Known limitation (flagged, not blocking)
`HUME_TTS_SAMPLE_RATE_HZ` is an assumed source sample rate (Hume's PCM docs don't specify it), and the binary-frame fix extends that same assumption to raw binary frames without independent confirmation. If wrong, audio would sound pitch-shifted rather than crash — worth verifying against Hume support before high call volume relies on it.

## Test plan
- [x] Full Hume/TTS/agent_runtime test suite passing locally (37+ tests across the files above)
- [x] Verified live against production data (Hume provider + 100 real voices with working preview URLs seeded directly, separately from this code deploy)
- [x] Reproduced and fixed the binary-frame crash against a real production call log
- [x] Deploy to a call-flow test environment and place a live test call through a Hume-configured agent

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -161,6 +161,8 @@ class TtsSettings(BaseModel):
     provider: str = Field(default="elevenlabs", validation_alias="TTS_PROVIDER")
     api_key: str = Field(default="", validation_alias="TTS_API_KEY")
     rime_api_key: str = Field(default="", validation_alias="RIME_API_KEY")
+    hume_api_key: str = Field(default="", validation_alias="HUME_API_KEY")
+    hume_sample_rate_hz: int = Field(default=48000, validation_alias="HUME_TTS_SAMPLE_RATE_HZ")
     elevenlabs_api_key: str = Field(default="", validation_alias="ELEVENLABS_API_KEY")
     elevenlabs_encryption_key: str = Field(
         default="", validation_alias="ELEVENLABS_ENCRYPTION_KEY"
@@ -464,6 +466,17 @@ class Settings(BaseSettings):
 
     # Rime Labs TTS Configuration
     RIME_API_KEY: str = ""
+
+    # Hume AI TTS Configuration
+    HUME_API_KEY: str = ""
+    # ASSUMED default — Hume's PCM streaming output sample rate is not
+    # documented publicly as of this integration (2026-08). 48000 Hz is a
+    # reasonable industry-standard guess for a modern neural TTS PCM stream,
+    # but MUST be verified against a real Hume account/API key before this
+    # goes to production traffic; override via env var if Hume support
+    # confirms a different value. Drives PCMStreamDownsampler's src_rate_hz
+    # when converting Hume's PCM16LE chunks down to mulaw 8kHz for Twilio.
+    HUME_TTS_SAMPLE_RATE_HZ: int = 48000
 
     # ElevenLabs Configuration
     ELEVENLABS_API_KEY: str = ""
@@ -1114,6 +1127,8 @@ class Settings(BaseSettings):
             provider=self.TTS_PROVIDER,
             api_key=self.TTS_API_KEY,
             rime_api_key=self.RIME_API_KEY,
+            hume_api_key=self.HUME_API_KEY,
+            hume_sample_rate_hz=self.HUME_TTS_SAMPLE_RATE_HZ,
             elevenlabs_api_key=self.ELEVENLABS_API_KEY,
             elevenlabs_encryption_key=self.ELEVENLABS_ENCRYPTION_KEY,
             enable_audio_tags=self.ENABLE_ELEVENLABS_AUDIO_TAGS,
--- a/app/core/secret_manager.py
+++ b/app/core/secret_manager.py
@@ -182,6 +182,43 @@ def get_rime_api_key() -> str:
 
 
 @lru_cache(maxsize=1)
+def get_hume_api_key() -> str:
+    """
+    Return the Hume AI TTS API key appropriate for the current environment.
+
+    - production/staging → GCP Secret Manager (secret name: HUME_API_KEY), with
+      env-var fallback so deployments that set the var directly still work.
+    - development → plain env-var / .env value.
+
+    Raises ValueError (development) or RuntimeError (staging/production) if no
+    key is available so callers fail at startup rather than sending unauthenticated
+    requests mid-call.
+    """
+    env = settings.ENVIRONMENT.lower()
+
+    if env in ("production", "staging"):
+        secret_key = _fetch_from_secret_manager("HUME_API_KEY")
+        if secret_key:
+            return secret_key
+        # Fall back to env var (covers deployments that inject the var via CI/CD).
+        env_key = getattr(settings, "HUME_API_KEY", "") or ""
+        if env_key.strip():
+            return env_key.strip()
+        raise RuntimeError(
+            f"HUME_API_KEY unavailable in {env}. "
+            "Set it in GCP Secret Manager (secret: HUME_API_KEY) or as an env var."
+        )
+
+    # development — plain env var / .env
+    env_key = getattr(settings, "HUME_API_KEY", "") or ""
+    if not env_key.strip():
+        raise ValueError(
+            "HUME_API_KEY is not set. Add it to your .env file for local development."
+        )
+    return env_key.strip()
+
+
+@lru_cache(maxsize=1)
 def _load_hubspot_production_credentials() -> Tuple[str, str]:
     """Fetch HubSpot OAuth app credentials from Secret Manager (cached after first call)."""
     client_id = _fetch_from_secret_manager("HUBSPOT_CLIENT_ID") or settings.HUBSPOT_CLIENT_ID
--- a/app/services/hume_tts_service.py
+++ b/app/services/hume_tts_service.py
@@ -1,0 +1,261 @@
+"""
+Hume AI TTS service — streaming WebSocket synthesis for real-time calling.
+
+API: wss://api.hume.ai/v0/tts/stream/input?api_key={api_key}&format_type=pcm&instant_mode=true
+Audio: server streams JSON messages containing base64-encoded raw PCM16LE
+mono audio chunks. There is no explicit per-chunk sample rate in the
+protocol; see `settings.HUME_TTS_SAMPLE_RATE_HZ` (app/core/config.py) for
+the configured-but-ASSUMED source rate this service downsamples from — that
+value is not confirmed by Hume's public docs as of this integration and
+should be verified against a real account/API key before production
+traffic depends on it.
+
+Telephony output: mulaw 8 kHz (matches Twilio MULAW 8000), produced
+incrementally via `PCMStreamDownsampler` as WebSocket chunks arrive.
+
+This integration deliberately mirrors Rime's "one WebSocket request per
+pipeline chunk" model (see rime_tts_service.py) — NOT a persistent,
+cross-turn streaming-input session (that's an ElevenLabs Phase 6-3 only
+feature, out of scope here).
+"""
+from __future__ import annotations
+
+import asyncio
+import base64
+import json
+import re
+import time
+from typing import Any, AsyncIterator
+
+from app.core.config import settings
+from app.core.logger import logger
+from app.core.secret_manager import get_hume_api_key
+from app.utils.audio_utils import PCMStreamDownsampler
+
+_HUME_WS_URL = "wss://api.hume.ai/v0/tts/stream/input"
+_DEFAULT_VOICE = "Male English Actor"
+_DEFAULT_VOICE_PROVIDER = "HUME_AI"
+
+# Mirrors the connect-timeout budget used elsewhere on the hot TTS path
+# (see app/services/elevenlabs_ws_session.py::_WS_CONNECT_TIMEOUT_S).
+_WS_CONNECT_TIMEOUT_S: float = 5.0
+# Idle read timeout per incoming message — guards against a stalled Hume
+# connection hanging a chunk forever.
+_WS_RECV_TIMEOUT_S: float = 15.0
+
+_UUID_RE = re.compile(
+    r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+)
+
+
+class HumeTtsServiceError(RuntimeError):
+    """Connect/auth/send/receive failure talking to Hume's TTS WebSocket."""
+
+
+def _looks_like_uuid(value: str) -> bool:
+    try:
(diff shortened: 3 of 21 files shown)


app/services/hume_tts_service.py
```diff
@@ -0,0 +1,284 @@
...
+            voice_external_id=voice_external_id,
+            voice_provider=voice_provider,
+            speed=speed,
+            description=description,
+        ):
+            chunks.append(chunk)
+        return b"".join(chunks)
+
+
+hume_tts_service = HumeTtsService()
```
codewithwaleed: 🔴 **Blocker: module-level singleton calls `get_hume_api_key()` at import time**

`HumeTtsService.__init__` calls `get_hume_api_key()`, which raises `ValueError` if `HUME_API_KEY` is not set. Any eager import of this module (before `conftest.py` seeds the env var, or in deployments where Hume isn't configured) will fail at import.

`rime_tts_service.py` avoids this by not calling `get_rime_api_key()` in `__init__`. Apply the same pattern:

```python
class HumeTtsService:
    def __init__(self) -> None:
        pass  # key resolved lazily on first call

    def _get_api_key(self) -> str:
        return get_hume_api_key()

    def _build_url(self) -> str:
        return f"{_HUME_WS_URL}?api_key={self._get_api_key()}&format_type=pcm&instant_mode=true"
```

app/core/config.py
```diff
@@ -465,6 +467,17 @@ class Settings(BaseSettings):
...
+    HUME_API_KEY: str = ""
+    # ASSUMED default — Hume's PCM streaming output sample rate is not
+    # documented publicly as of this integration (2026-08). 48000 Hz is a
+    # reasonable industry-standard guess for a modern neural TTS PCM stream,
+    # but MUST be verified against a real Hume account/API key before this
+    # goes to production traffic; override via env var if Hume support
+    # confirms a different value. Drives PCMStreamDownsampler's src_rate_hz
+    # when converting Hume's PCM16LE chunks down to mulaw 8kHz for Twilio.
+    HUME_TTS_SAMPLE_RATE_HZ: int = 48000
+
```
codewithwaleed: 🟠 **Major: unverified `HUME_TTS_SAMPLE_RATE_HZ` — wrong value produces silently garbled audio on every Twilio call**

The comment says "ASSUMED default — not confirmed by Hume's public docs". If the actual stream rate differs from 48000 Hz, `PCMStreamDownsampler` downsamples at the wrong ratio and output is pitch-shifted and distorted with no runtime error.

Add a startup log warning so operators know to verify before enabling in production:

```python
# in app/main.py lifespan, after get_hume_api_key() succeeds:
logger.warning(
    "Hume TTS sample rate assumed %d Hz (unverified) — "
    "confirm against a real Hume account before production traffic. "
    "Override via HUME_TTS_SAMPLE_RATE_HZ if different.",
    settings.HUME_TTS_SAMPLE_RATE_HZ,
)
```

app/services/hume_tts_service.py
```diff
@@ -0,0 +1,284 @@
...
+
+        voice_payload = _build_voice_payload(voice_external_id, voice_provider)
+        message: dict[str, Any] = {
+            "text": text.strip(),
+            "voice": voice_payload,
+            "speed": float(speed),
+            "close": True,
+        }
+        if description:
+            message["description"] = description[:100]
```
codewithwaleed: 🟠 **Major: `description` is silently truncated to 100 characters — unverified limit**

`description[:100]` may cut a long acting instruction mid-sentence. Hume's public docs don't state a 100-char limit as of this integration. Either document the verified limit or let Hume's API reject oversized values:

```python
if description:
    if len(description) > 100:
        logger.debug("[Hume] description truncated from %d to 100 chars", len(description))
    message["description"] = description[:100]
# OR: remove the truncation and let the API validate
```

app/utils/tts_adapter.py
```diff
@@ -465,6 +519,128 @@ def stream_synthesize(
...
+            return payload.get("voices_page") or payload.get("voices") or []
+        if isinstance(payload, list):
+            return payload
+        return []
+
+    def normalize_voice_payload(self, payload: dict[str, Any]) -> dict[str, Any]:
+        return {
+            "external_voice_id": payload.get("id") or payload.get("name"),
+            "display_name": payload.get("name") or "Hume Voice",
+            "language_code": None,
```
codewithwaleed: 🟡 **Minor: default voice string duplicated in 4 places — use an importable constant**

`"Male English Actor"` appears in `hume_tts_service.py`, `agent_runtime.py`, `tts_stream_mixin.py` (×2), and `livekit_browser_call_handler.py`. A typo in one copy would cause a silent mismatch.

Export a single constant from `hume_tts_service.py`:
```python
# hume_tts_service.py
HUME_DEFAULT_VOICE = "Male English Actor"
```
Then import it in the other files instead of repeating the string literal.

app/core/secret_manager.py
```diff
@@ -181,6 +181,43 @@ def get_rime_api_key() -> str:
...
+            "Set it in GCP Secret Manager (secret: HUME_API_KEY) or as an env var."
+        )
+
+    # development — plain env var / .env
+    env_key = getattr(settings, "HUME_API_KEY", "") or ""
+    if not env_key.strip():
+        raise ValueError(
+            "HUME_API_KEY is not set. Add it to your .env file for local development."
+        )
+    return env_key.strip()
```
codewithwaleed: 🟡 **Minor: `get_hume_api_key` is decorated with `@lru_cache(maxsize=1)` — the cache is not cleared between tests**

`get_rime_api_key` has the same issue. If a test modifies `HUME_API_KEY` mid-run, the cached value from a prior call will be returned. Add a `cache_clear()` call in a test fixture:

```python
# in tests/core/test_secret_manager_hume.py
@pytest.fixture(autouse=True)
def clear_hume_key_cache():
    from app.core.secret_manager import get_hume_api_key
    get_hume_api_key.cache_clear()
    yield
    get_hume_api_key.cache_clear()
```

app/utils/audio_utils.py
```diff
@@ -358,6 +358,104 @@ def downsample_linear_samples(samples: list[int], src_rate_hz: int, dst_rate_hz:
...
+    - a one-time WAV header at the start of the stream
+    - src -> dst box-average downsampling
+    - direct mu-law encoding of the resulting samples (via linear_to_ulaw_sample)
+
+    `src_rate_hz` must be an integer multiple of `dst_rate_hz` (matches the
+    existing PCM16KStreamDownsampler's assumption — no fractional-ratio
+    resampling here).
+    """
+
+    __slots__ = ("_buf", "_header_done", "_factor", "_bytes_per_group")
```
codewithwaleed: 🔵 **Nit: `PCMStreamDownsampler` validates `src_rate_hz % dst_rate_hz != 0` but the error message says "Unsupported resample ratio" without listing valid alternatives**

For a future maintainer, a more helpful error would indicate the constraint:

```python
raise ValueError(
    f"src_rate_hz must be a positive integer multiple of dst_rate_hz "
    f"(got {src_rate_hz} -> {dst_rate_hz})"
)
```

codewithwaleed (changes requested):
## PR #212 Review — Add Hume AI as a real-time streaming TTS provider

### Summary
This PR integrates Hume AI TTS via a WebSocket streaming endpoint, following the same patterns established by the Rime integration: a dedicated `HumeSyncEventLoopBridge` for sync callers, an `async_stream_synthesize` path for the real-time voice hot path, a `PCMStreamDownsampler` that generalises the existing `PCM16KStreamDownsampler`, and proper startup key validation. The architecture is sound and consistent with the existing provider patterns. A few issues need attention before merge.

---

### 🔴 Blocker

**1. `HumeTtsService` is constructed as a module-level singleton at import time — raises `ValueError` in every test environment that imports `hume_tts_service.py`, even indirectly**

```python
# app/services/hume_tts_service.py (last line)
hume_tts_service = HumeTtsService()  # calls get_hume_api_key() at import time
```

`HumeTtsService.__init__` calls `get_hume_api_key()`, which raises `ValueError` if `HUME_API_KEY` is not set. `conftest.py` sets the key via `os.environ.setdefault` — but only *after* any module that imports `hume_tts_service` at the top level (not lazily) would already have triggered the import. `HumeTTSAdapter.synthesize` and `async_stream_synthesize` both guard this with a local `from app.services.hume_tts_service import hume_tts_service` import, which is correct. But if anything else imports the module eagerly (a router, a service initialised at startup, a test that imports `tts_adapter` before conftest runs), it will fail.

The established pattern in this codebase (`rime_tts_service.py`) avoids this by *not* calling `get_rime_api_key()` in `__init__` and instead resolving it lazily. Apply the same fix:

```python
class HumeTtsService:
    def __init__(self) -> None:
        pass  # key resolved lazily on first call

    def _get_api_key(self) -> str:
        return get_hume_api_key()

    def _build_url(self) -> str:
        return f"{_HUME_WS_URL}?api_key={self._get_api_key()}&format_type=pcm&instant_mode=true"
```

---

### 🟠 Major Issues

**2. `HUME_TTS_SAMPLE_RATE_HZ` is documented as an unverified assumption — if wrong, every Twilio call using Hume will produce garbled audio silently**

The config comment explicitly says "ASSUMED default — not confirmed by Hume's public docs". If the actual stream rate differs from 48000 Hz, `PCMStreamDownsampler` will downsample at the wrong ratio and the output will be pitch-shifted and distorted. There is no runtime check or log warning that would surface this — the call simply sounds wrong.

At minimum, log a prominent startup warning so operators know to verify before enabling Hume in production:

```python
# in lifespan, after get_hume_api_key():
logger.warning(
    "Hume TTS sample rate is assumed to be %d Hz (unverified) — "
    "confirm against a real Hume account before production traffic. "
    "Override via HUME_TTS_SAMPLE_RATE_HZ env var if needed.",
    settings.HUME_TTS_SAMPLE_RATE_HZ,
)
```

**3. `PCMStreamDownsampler._drain` uses integer division by `factor` but falls back to `n` (actual group size) in the `take_all` branch — this produces a different average for the final partial group**

```python
avg = int(total / (factor if not take_all else n))
```

In the `take_all=True` (flush) path, `n` is the actual number of samples in the partial group (smaller than `factor`), so this correctly averages the partial group's samples. But in the `take_all=False` path, `n` might be *smaller* than `factor` on the last iteration if `usable` is not a perfect multiple of `group` — which should not happen since `usable -= usable % group` ensures it is. However, using `n` for normalization in `take_all=True` is correct only if `n * 2 == end - i`. Verify the edge case where `usable == 2` (single sample, `i=0`, `end=2`, `n=1`, `factor=6` for 48→8 ratio) — the result is `int(sample / 1)` which is correct (the sample itself, not divided by factor).

This is correct as written. No change needed — adding this to confirm the logic is sound.

**4. `description` in the WS payload is truncated to 100 characters (`description[:100]`) with no warning — a long acting instruction might be silently truncated mid-sentence**

Hume's API may not impose a 100-char limit publicly. If the limit is wrong or the API accepts longer values, truncating at 100 would produce instructions like "Speak in a warm, friendly tone that conveys empathy and..." instead of the full prompt. Either verify the limit and document it, or don't truncate (let Hume's API validate length):

```python
# Option A: document verified limit
if description:
    if len(description) > 100:
        logger.debug("[Hume] description truncated from %d to 100 chars", len(description))
    message["description"] = description[:100]

# Option B: no truncation — let Hume's API reject oversized values
if description:
    message["description"] = description
```

---

### 🟡 Minor Issues

**5. `_redact` in `hume_tts_service.py` re-applies on every exception — it's protective but the `websockets` library today doesn't embed the request URI in exception messages**

The comment acknowledges this is "defensive insurance against a future `websockets` version". This is fine — no change needed, but it's worth noting that the URL is constructed via `_build_url()` which embeds the raw API key, so if that URL string ever appears in logs/exceptions, `_redact` correctly handles it. ✓

**6. Default voice "Male English Actor" is hardcoded in 4 separate places** — `hume_tts_service.py`, `agent_runtime.py`, `tts_stream_mixin.py` (twice), and `livekit_browser_call_handler.py`. A constant in `hume_tts_service.py` is exported as `_DEFAULT_VOICE` but not imported by the other callsites. Extract to a single importable constant:

```python
# hume_tts_service.py
HUME_DEFAULT_VOICE = "Male English Actor"  # public — importable by agent_runtime, tts_stream_mixin
```

**7. `HumeTTSAdapter.list_voices` uses a sync `requests.get()` call** — the comment says "Admin voice-picker UI path only — latency doesn't matter here". This is consistent with how `RimeTTSAdapter.list_voices` works. ✓ No change needed.

---

### Missing Test Coverage

- **`get_hume_api_key()` with `lru_cache`**: the cache is not cleared between tests — any test that calls `get_hume_api_key()` with a specific key set will cache that value for all subsequent tests in the same process. `test_secret_manager_hume.py` should call `get_hume_api_key.cache_clear()` in a fixture.
- **`PCMStreamDownsampler` with 48000→8000 ratio** (factor=6): the existing test suite covers `feed`/`flush` but the ratio in tests may not match the actual production ratio used for Hume. A parametrized test with `src_rate=48000, dst_rate=8000` covering the WAV-header strip path would confirm the production path works.

---

### Recommendation: **Request Changes** — fix the eager module-level `HumeTtsService()` singleton construction (#1) before merge; add the startup warning for the unverified sample rate (#2).

PR #213, opened 2026-08-19
feat(voice): add xAI as a real-time streaming TTS provider

Aliict:
## Summary
- Integrates xAI's WebSocket TTS (`wss://api.x.ai/v1/tts`) as a 5th real-time streaming TTS provider, alongside ElevenLabs, Google, Rime, and Hume — reusing the existing multi-provider TTS pipeline architecture (`app/utils/tts_adapter.py`, `tts_provider_capabilities.py`, `tts_catalog_service.py`, `TtsProviderEnum`).
- Requests `codec=mulaw&sample_rate=8000` directly from xAI, so no local resampling is needed downstream (unlike Hume, which only streams PCM).
- One WebSocket per synthesis chunk, matching the Rime/Hume connection model; cancellation/barge-in uses the existing generic task-cancel → WS-close path.
- Registers the provider everywhere the other providers are wired in: adapter registry, capabilities, voice catalog seeding, `TtsProviderEnum`, OpenAPI spec, and default-voice wiring for both the Twilio and browser/LiveKit call transports.
- Reads `XAI_API_KEY` via `settings.XAI_API_KEY` directly (matching this repo's existing xAI STT integration), not via `secret_manager.py`'s GCP-backed path used by Rime/Hume — a deliberate, documented tradeoff (weaker secret-management story in staging/prod for this one provider) worth a conscious sign-off, not an oversight.

## Test plan
- [x] 29 new tests in `tests/voice/test_xai_tts_and_barge_in.py` covering streaming/chunking, connection URL + Bearer auth header, `audio.delta`/`audio.done`/`error` message handling, mulaw/8kHz format, cancellation/barge-in, timeouts/disconnects, empty/malformed/binary-frame edge cases, and provider registration.
- [x] Full TTS-related regression sweep (`tests/voice/ tests/services/ tests/core/ tests/utils/ -k "tts or xai"`): 268 passed, 0 regressions to ElevenLabs/Google/Rime/Hume.
- [x] `ruff check` clean on all changed files.
- [x] Real smoke test against live `wss://api.x.ai/v1/tts` with the configured `XAI_API_KEY`: connection succeeded, ~1870ms first-audio-chunk latency, 8 chunks / ~4.87s of mulaw/8kHz audio, clean `audio.done`.
- [x] Message protocol (field names, `error`/`audio.done` handling) verified line-by-line against xAI's official docs in a second review pass — no protocol-mismatch bugs (the class of bug that hit the earlier Hume integration).
- [x] `docs-vault-sync` (adds xAI to the vault's TTS provider notes) — not yet run, can follow post-merge.

### DIFF
--- a/app/services/xai_tts_service.py
+++ b/app/services/xai_tts_service.py
@@ -74,12 +74,17 @@ class XaiTtsServiceError(RuntimeError):
 
 
 def _mask_key(key: str | None) -> str:
-    """Safe, non-reversible preview for diagnostics — never logs the raw key."""
+    """Safe, non-reversible preview for diagnostics — never logs the raw key.
+
+    Only a 4-char prefix is shown (no suffix) — a prefix+suffix reveal on a
+    short key can leak a large fraction of it (flagged in PR #207 for the
+    equivalent AssemblyAI helper).
+    """
     if not key:
         return "<empty>"
     if len(key) <= 8:
         return f"<masked len={len(key)}>"
-    return f"{key[:4]}...{key[-2:]} (len={len(key)})"
+    return f"{key[:4]}{'*' * min(8, len(key) - 4)} (len={len(key)})"
 
 
 class XaiTtsService:
@@ -164,7 +169,8 @@ class XaiTtsService:
             # trace in the logs if that upstream invariant ever breaks.
             logger.warning(
                 "[xAI TTS] text.delta exceeds %d chars (got %d) — truncating",
-                _MAX_TEXT_DELTA_CHARS, len(stripped_text),
+                _MAX_TEXT_DELTA_CHARS,
+                len(stripped_text),
             )
             stripped_text = stripped_text[:_MAX_TEXT_DELTA_CHARS]
 
--- a/app/services/xai_tts_service.py
+++ b/app/services/xai_tts_service.py
@@ -1,0 +1,314 @@
+"""
+xAI Grok TTS service — streaming WebSocket synthesis for real-time calling.
+
+API: wss://api.x.ai/v1/tts
+Auth: Bearer token via the `Authorization` header on the WebSocket upgrade
+request (matches how `app/services/xai_grok_stt_service.py` authenticates
+xAI's STT WebSocket — xAI's TTS endpoint does not document query-string
+auth). Query params configure the session: `language` (required, BCP-47 or
+`auto`), `voice` (default `eve`), `codec` (`mp3`/`wav`/`pcm`/`mulaw`/`ulaw`/
+`alaw`), `sample_rate`, `bit_rate` (mp3 only), `speed` (0.7-1.5),
+`optimize_streaming_latency` (0=off, 1=moderate, 2=aggressive),
+`text_normalization`, `with_timestamps`.
+
+Telephony output: this service always requests `codec=mulaw&sample_rate=8000`
+(matches Twilio MULAW 8000) directly from xAI — unlike Hume (PCM-only,
+requires `PCMStreamDownsampler` to get to mulaw/8kHz), xAI natively emits
+telephony-ready audio, so no resampling step exists in this file.
+
+Client->server JSON messages: `text.delta` (`{"type":"text.delta","delta":
+"..."}`, incremental text, <=15k chars), `text.done` (flush + synthesize the
+accumulated text), `text.clear` (cancel the current utterance without closing
+the connection — xAI's documented mechanism for a persistent, cross-turn
+session). Server->client JSON messages: `audio.delta` (base64 audio chunk),
+`audio.done` (utterance complete, connection may be reused), `audio.clear`
+(ack of `text.clear`), `error` (connection stays open per xAI's docs).
+
+Connection model: this integration deliberately mirrors Hume/Rime's "one
+WebSocket per synthesis chunk" model (open -> text.delta(s) + text.done ->
+drain audio.delta until audio.done -> close) — NOT a persistent, cross-turn
+streaming session (that's ElevenLabs' Phase 6-3-only
+`elevenlabs_ws_session.py` pattern, out of scope here, same as Hume's own
+docstring scopes itself). This is consistent with how cancellation already
+works elsewhere in this codebase: the pipeline layer cancels the consuming
+async generator/task on barge-in, which propagates through `await ws.recv()`
+and closes the socket in a `finally` block below — that alone is a valid
+stop-synthesis mechanism per xAI's docs; nothing requires sending
+`text.clear` before closing. `text.clear`/`audio.clear` are consequently
+unused here but are xAI's documented mechanism for a future persistent-
+session mode, should this integration ever grow one.
+"""
+from __future__ import annotations
+
+import asyncio
+import base64
+import json
+import time
+from typing import Any, AsyncIterator
+from urllib.parse import urlencode
+
+from app.core.config import settings
+from app.core.logger import logger
+
+_XAI_TTS_URL = "wss://api.x.ai/v1/tts"
+# Public — imported by agent_runtime.py / tts_stream_mixin.py /
+# livekit_browser_call_handler.py so the default voice string exists in
+# exactly one place, mirroring HUME_DEFAULT_VOICE's rationale.
+XAI_DEFAULT_VOICE = "eve"
+_DEFAULT_VOICE = XAI_DEFAULT_VOICE
+_DEFAULT_LANGUAGE = "auto"
+
+# Mirrors the connect-timeout budget used elsewhere on the hot TTS path
+# (see app/services/hume_tts_service.py / elevenlabs_ws_session.py).
+_WS_CONNECT_TIMEOUT_S: float = 5.0
+# Idle read timeout per incoming message — guards against a stalled xAI
+# connection hanging a chunk forever.
+_WS_RECV_TIMEOUT_S: float = 15.0
+
+# xAI's documented ceiling on a single text.delta payload.
+_MAX_TEXT_DELTA_CHARS = 15_000
+
+
+class XaiTtsServiceError(RuntimeError):
+    """Connect/auth/send/receive failure talking to xAI's TTS WebSocket."""
+
+
+def _mask_key(key: str | None) -> str:
+    """Safe, non-reversible preview for diagnostics — never logs the raw key."""
+    if not key:
+        return "<empty>"
+    if len(key) <= 8:
+        return f"<masked len={len(key)}>"
+    return f"{key[:4]}...{key[-2:]} (len={len(key)})"
+
+
+class XaiTtsService:
+    """Thin async wrapper around the xAI TTS WebSocket streaming endpoint."""
+
+    def __init__(self) -> None:
+        # Resolved lazily on first call, not here -- this class is
+        # constructed as a module-level singleton at import time, and xAI
+        # TTS is an optional provider most tenants never configure. Eagerly
+        # reading/validating settings.XAI_API_KEY here would surface a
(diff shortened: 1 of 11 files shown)


app/utils/tts_adapter.py
```diff
@@ -139,6 +140,60 @@ def run(self, coro: "Coroutine[Any, Any, bytes]") -> bytes:
...
+
+    def __init__(self) -> None:
+        self._loop: asyncio.AbstractEventLoop | None = None
+        self._start_lock = threading.Lock()
+
+    def _ensure_started(self) -> asyncio.AbstractEventLoop:
+        loop = self._loop
+        if loop is not None:
+            return loop
+        with self._start_lock:
```
codewithwaleed: 🟠 **Major: `self._api_key` is stored but never used — dead code with a subtle inconsistency**

Every synthesis call delegates to `xai_tts_service.synthesize()` which calls `_get_api_key()` and re-reads `settings.XAI_API_KEY` itself. The adapter's `self._api_key` is never passed to the service.

If `XAI_API_KEY` is present at adapter construction but later removed (env refresh, test teardown), adapter construction passes but synthesis raises `XaiTtsServiceError` — a different exception type than the `ValueError` from the fail-fast check here.

Fix — validate without storing:
```python
if not (settings.XAI_API_KEY or "").strip():
    raise ValueError(
        "XAI_API_KEY is not set. Add it to your environment/.env to use xAI TTS."
    )
# Do NOT assign self._api_key — service re-reads it via _get_api_key()
```

app/utils/tts_adapter.py
```diff
@@ -642,6 +697,118 @@ def stream_synthesize(
...
+    def list_voices(self) -> list[dict[str, Any]]:
+        # xAI's GET /v1/tts/voices exists but there is no vendor SDK / prior
+        # verified integration for it in this codebase — return a small
+        # static list (matches Rime's approach) so the admin voice-picker UI
+        # has something to show without depending on an unverified live call.
+        return [
+            {"voice_id": "eve", "name": "Eve"},
+        ]
+
+    def normalize_voice_payload(self, payload: dict[str, Any]) -> dict[str, Any]:
```
codewithwaleed: 🟠 **Major: `stream_synthesize` raises `NotImplementedError` with no construction-time guard**

`get_tts_adapter("xai")` constructs `XaiTTSAdapter()` without any indication that `stream_synthesize` is not available. Any caller that does `adapter.stream_synthesize(...)` (e.g., a generic mixin fallback or a future test) will hit this at runtime with no early warning.

The `TTSProviderCapabilities` entry for `"xai"` sets `supports_streaming=True`, which is correct for `async_stream_synthesize`, but a caller reading only the capability flags has no way to know `stream_synthesize` will raise.

Options:
1. Add a `supports_async_streaming_only: bool` capability flag and gate call sites on it.
2. Implement a `stream_synthesize` wrapper that calls `asyncio.run(self.synthesize(...))` (safe when called from a sync context with no running loop — matches the bridge pattern used for Rime/Hume's sync paths).
3. At minimum, log a warning before raising so the exception traces back to the right place.

app/services/xai_tts_service.py
```diff
@@ -0,0 +1,314 @@
...
+                    raise
+                except asyncio.TimeoutError as exc:
+                    raise XaiTtsServiceError(
+                        f"xAI TTS WebSocket idle timeout after {_WS_RECV_TIMEOUT_S}s"
+                    ) from exc
+                except Exception as exc:
+                    # ConnectionClosed (normal or abnormal) and any other
+                    # protocol-level failure land here. Unlike Hume, xAI's
+                    # documented end-of-utterance signal is the in-band
+                    # `audio.done` JSON message, not a socket close — so a
```
codewithwaleed: 🟡 **Minor: `_mask_key` leaks 6 characters of the API key (`key[:4]` + `key[-2:]`)**

Same issue flagged in the AssemblyAI PR (#207). For a 10-12 char key, showing 4+2=6 chars is a significant fraction. Use only the 4-char prefix:
```python
def _mask_key(key: str | None) -> str:
    if not key:
        return "<empty>"
    if len(key) <= 8:
        return f"<masked len={len(key)}>"
    return f"{key[:4]}{'*' * min(8, len(key) - 4)} (len={len(key)})"
```

app/utils/tts_adapter.py
```diff
@@ -139,6 +140,60 @@ def run(self, coro: "Coroutine[Any, Any, bytes]") -> bytes:
...
+            if self._loop is not None:
+                return self._loop
+
+            ready = threading.Event()
+            state: dict[str, asyncio.AbstractEventLoop] = {}
+
+            def _run_forever() -> None:
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
+                state["loop"] = loop
```
codewithwaleed: 🟡 **Minor: `list_voices()` returns only `"eve"` — admins can't discover other xAI voices via the UI**

xAI's TTS API documents multiple voices and a `GET /v1/tts/voices` endpoint. The static stub means the admin voice-picker shows only "eve" regardless of what the tenant's account can access.

Since this matches Rime's approach (acknowledged in the comment), it's an acceptable starting point — but add a comment linking to xAI's voice list so admins know what IDs to type:
```python
# xAI's known voices (as of Aug 2026): eve, etc.
# See https://docs.x.ai/docs/guides/text-to-speech for the current list.
# A live GET /v1/tts/voices fetch is not wired yet — add it if the voice
# catalog needs to stay current without manual updates.
return [
    {"voice_id": "eve", "name": "Eve"},
]
```

tests/voice/test_xai_tts_and_barge_in.py
```diff
@@ -0,0 +1,476 @@
...
+        from app.services.tts_catalog_service import TTSCatalogService
+
+        svc = TTSCatalogService()
+        # ensure_default_provider() is DB-backed; assert against the static
+        # seed spec list directly instead, mirroring how this module has no
+        # existing unit test hitting a real DB.
+        import inspect
+
+        source = inspect.getsource(svc.ensure_default_provider)
+        assert '"slug": "xai"' in source
```
codewithwaleed: 🟡 **Minor: `inspect.getsource` to assert a string literal is a fragile test pattern**

```python
source = inspect.getsource(svc.ensure_default_provider)
assert '"slug": "xai"' in source
```

This breaks silently on any Black reformatting, quote-style change, or method rename. It tests the text of the source, not the behavior.

Better: extract the seed list as a class-level constant and assert against the Python list directly:
```python
# In tts_catalog_service.py
SEED_PROVIDERS = [
    {"slug": "rime", ...},
    {"slug": "hume", ...},
    {"slug": "xai", ...},
]

# In the test
from app.services.tts_catalog_service import TTSCatalogService
slugs = [p["slug"] for p in TTSCatalogService.SEED_PROVIDERS]
assert "xai" in slugs
```

Or at minimum use `any(p["slug"] == "xai" for p in svc._seed_list)` rather than source inspection.

codewithwaleed (changes requested):
## PR #213 Review — xAI Grok TTS provider

### Summary
This PR correctly follows the Hume/Rime provider pattern: `XaiTtsService` for the WebSocket layer, `XaiTTSAdapter` for the adapter interface, `_XaiSyncEventLoopBridge` for the sync-bridge path, and a capability entry in `TTSProviderCapabilities`. The telephony-native mulaw/8kHz output (no resampler needed, unlike Hume) and the one-shot WebSocket connection model are well-motivated and clearly documented.

A few issues require attention before merge.

---

### 🟠 Major Issues

**1. `XaiTTSAdapter.__init__` reads and stores `self._api_key` but the field is never used**

The adapter's constructor eagerly reads `settings.XAI_API_KEY` and stores it on `self._api_key` for fail-fast validation — but every synthesis call delegates to `xai_tts_service.synthesize()` / `xai_tts_service.stream_text_to_speech()`, which call `self._get_api_key()` internally and re-read `settings.XAI_API_KEY` themselves. The adapter's `self._api_key` is never passed to the service and is dead code.

This creates a subtle inconsistency: if `XAI_API_KEY` is set at adapter construction time but removed from settings afterward (env refresh, test isolation), the adapter passes its own fail-fast but the service will then raise `XaiTtsServiceError` mid-call, which surfaces as a different exception type than the `ValueError` from adapter construction.

Fix — either pass the key explicitly to the service, or remove the storage and just do the check without assigning:
```python
def __init__(self) -> None:
    from app.services.xai_tts_service import XAI_DEFAULT_VOICE
    self._DEFAULT_VOICE = XAI_DEFAULT_VOICE
    # Fail fast without storing; synthesis path re-reads via xai_tts_service._get_api_key()
    if not (settings.XAI_API_KEY or "").strip():
        raise ValueError(
            "XAI_API_KEY is not set. Add it to your environment/.env to use xAI TTS."
        )
```

**2. `XaiTTSAdapter.stream_synthesize` raises `NotImplementedError` — not guarded at registration**

`BaseTTSProviderAdapter` requires `stream_synthesize` as an abstract method, and the implementation raises `NotImplementedError`. The comment says "callers should use `async_stream_synthesize`" — but if any code path (e.g., a sync fallback in `tts_stream_mixin.py`, or a test that calls `stream_synthesize` via the adapter interface) hits this at runtime, it raises an unhandled exception with no warning at construction or registration time.

`get_tts_adapter("xai")` constructs a new `XaiTTSAdapter()` on every call with no indication that `stream_synthesize` is unavailable. The `NotImplementedError` should at minimum be documented on the `BaseTTSProviderAdapter` interface, and `tts_stream_mixin.py` should explicitly guard against calling `stream_synthesize` on providers where `capabilities.supports_streaming is True` but `stream_synthesize` raises — which is a contradiction in the current capability flags.

Consider checking `hasattr(adapter, 'async_stream_synthesize')` before calling `adapter.stream_synthesize` in the mixin, or use a separate capability flag like `supports_async_streaming_only`.

---

### 🟡 Minor Issues

**3. `_mask_key` leaks 6 characters of the API key (`key[:4]` + `key[-2:]`)**

Same issue flagged in PR #207 (AssemblyAI service). For a short key, 4+2=6 chars is a significant fraction. The pattern from other secret-manager functions uses only a 4-char prefix. Fix:
```python
return f"{key[:4]}{'*' * min(8, len(key) - 4)} (len={len(key)})"
```

**4. `list_voices()` returns only a static `[{"voice_id": "eve", "name": "Eve"}]` stub**

xAI documents `GET /v1/tts/voices` (equivalent path for the REST API). The admin voice-picker will show only "eve" regardless of what voices the tenant's xAI account actually has access to — they can type any voice ID but never discover others via the UI. The comment acknowledges this ("matches Rime's approach") but Rime's static list has more than one voice. At minimum, add a comment that links to xAI's voice list so admins know what IDs are valid:
```python
# xAI's documented voices as of 2026-08: eve, etc.
# See https://docs.x.ai/docs/guides/text-to-speech for the current list.
# A live /v1/tts/voices fetch is not wired up yet — add it if the voice
# catalog needs to stay current without manual updates.
```

**5. `test_catalog_seed_includes_xai` uses `inspect.getsource` to assert a string literal**

```python
source = inspect.getsource(svc.ensure_default_provider)
assert '"slug": "xai"' in source
```

This assertion breaks on any formatting change (different quote style, whitespace, Black reformatting). The actual thing to test is that a DB-seeded provider with `slug == "xai"` exists — either mock the DB session and assert the ORM call, or extract the seed list as a class constant and assert against it directly. As-is, the test is testing the source text of the method, not its behavior.

---

### Recommendation: **Request Changes** — fix the dead `self._api_key` field (#1) and document the `stream_synthesize` / `NotImplementedError` risk more explicitly before merge (#2).

PR #214, opened 2026-08-20
feat(integrations): add Slack OAuth integration for post-call summaries

Aliict:
## Summary
- Adds a Slack OAuth integration for delivering post-call summaries to a Slack channel, following the existing HubSpot/Salesforce OAuth pattern (`WorkspaceIntegration` table, AES-256-GCM token encryption, signed-JWT OAuth state).
- New `app/routers/slack_integration.py` (mounted at `/api/v1/integrations/slack`): connect (dual redirect/JSON), callback, disconnect, status, channel listing, default-channel selection.
- New `CallFlow` columns (`slack_summary_enabled`, `slack_channel_id`, `slack_channel_name`) — Slack is **off by default** per call flow, matching the existing `email_summary_enabled` opt-in pattern. Exposed via the existing `PUT /api/v2/flows/{flow_id}/post-call-actions-settings` endpoint (no new endpoint added).
- New post-call hook in `CallSessionService.update_call_session_status()`: fires only when both the call flow has opted in and the tenant has Slack connected; entirely fail-open (wrapped in try/except that only logs) so a Slack outage or misconfiguration never affects call completion.
- OAuth scope includes `chat:write.public` (verified against Slack's own API docs) — without it, posting to public channels the bot wasn't manually invited to would silently fail.

## Test plan
- [x] `tests/services/test_slack_service.py` (56 tests) — OAuth state/CSRF, token encryption round-trip, channel listing (raises on failure — admin action), send-summary fail-open behavior, tenant isolation (real two-tenant DB test), disconnect/reconnect, async dispatch pattern.
- [x] `tests/api/test_slack_integration.py` (25 tests) — RBAC per endpoint, connect redirect vs JSON mode, callback state verification, error handling.
- [x] Updated `tests/api/v2/test_post_call_actions_settings.py` and `tests/services/test_call_session_post_call_email_hook.py` for the new Slack fields/hook.
- [x] Full existing suite (`pytest tests/ --ignore=tests/integration`) passes: 2984 passed, 1 pre-existing unrelated flaky timing test deselected (`test_audio_transcoder.py`, not touched by this branch).
- [x] `ruff check` / `black --check` clean on all touched files.
- [x] Migration `c4e70c96cb81` applied cleanly (pure additive `ADD COLUMN`, safe on a live table).
- [x] Reviewed by `code-reviewer` agent — no blocking findings; one test-quality gap (mocked tenant-isolation test) fixed with a real two-tenant DB test.
- [x] Manually cross-checked OAuth/API implementation against `docs.slack.dev` (oauth.v2.access, conversations.list, chat.postMessage) — found and fixed a missing `chat:write.public` scope.

## Setup required post-merge
A Slack app needs to be created with scopes `channels:read`, `groups:read`, `chat:write`, `chat:write.public`, and redirect URL `{WEBHOOK_BASE_URL}/api/v1/integrations/slack/callback`. Env vars: `SLACK_CLIENT_ID`, `SLACK_CLIENT_SECRET`, `SLACK_TOKEN_ENCRYPTION_KEY` (optionally `SLACK_REDIRECT_URI`) — added to `.env.example` and `CLAUDE.md`.

https://claude.ai/code/session_01Y3RE1KRZwQZPwzHTypz34V

### DIFF
--- a/app/core/db_encryption.py
+++ b/app/core/db_encryption.py
@@ -253,7 +253,9 @@ def _hubspot_aes_key() -> bytes:
     return hashlib.sha256(key.encode("utf-8")).digest()
 
 
-def encrypt_hubspot_token(plaintext: str, db: Session) -> str:  # noqa: ARG001 - db kept for call-site parity
+def encrypt_hubspot_token(
+    plaintext: str, db: Session
+) -> str:  # noqa: ARG001 - db kept for call-site parity
     """Encrypt *plaintext* HubSpot OAuth token with AES-256-GCM.
 
     Performed in Python via ``cryptography`` (not pgcrypto SQL — OpenPGP symmetric
@@ -286,7 +288,7 @@ def decrypt_hubspot_token(ciphertext: str, db: Session) -> str:
     if ciphertext.startswith(_HUBSPOT_AESGCM_PREFIX):
         key = _hubspot_aes_key()
         try:
-            raw = base64.b64decode(ciphertext[len(_HUBSPOT_AESGCM_PREFIX):])
+            raw = base64.b64decode(ciphertext[len(_HUBSPOT_AESGCM_PREFIX) :])
             nonce, body = raw[:12], raw[12:]
             return AESGCM(key).decrypt(nonce, body, None).decode("utf-8")
         except Exception as exc:
@@ -332,7 +334,9 @@ def _salesforce_aes_key() -> bytes:
     return hashlib.sha256(key.encode("utf-8")).digest()
 
 
-def encrypt_salesforce_token(plaintext: str, db: Session) -> str:  # noqa: ARG001 - db kept for call-site parity
+def encrypt_salesforce_token(
+    plaintext: str, db: Session
+) -> str:  # noqa: ARG001 - db kept for call-site parity
     """Encrypt *plaintext* Salesforce OAuth token with AES-256-GCM.
 
     Performed in Python via ``cryptography`` (not pgcrypto SQL — OpenPGP symmetric
@@ -344,10 +348,14 @@ def encrypt_salesforce_token(plaintext: str, db: Session) -> str:  # noqa: ARG00
     key = _salesforce_aes_key()
     nonce = os.urandom(12)  # 96-bit nonce, the standard size for AES-GCM
     ciphertext = AESGCM(key).encrypt(nonce, plaintext.encode("utf-8"), None)
-    return _SALESFORCE_AESGCM_PREFIX + base64.b64encode(nonce + ciphertext).decode("ascii")
+    return _SALESFORCE_AESGCM_PREFIX + base64.b64encode(nonce + ciphertext).decode(
+        "ascii"
+    )
 
 
-def decrypt_salesforce_token(ciphertext: str, db: Session) -> str:  # noqa: ARG001 - db kept for call-site parity
+def decrypt_salesforce_token(
+    ciphertext: str, db: Session
+) -> str:  # noqa: ARG001 - db kept for call-site parity
     """Decrypt a Salesforce OAuth token written by :func:`encrypt_salesforce_token`.
 
     Raises ``ValueError`` if ``SALESFORCE_TOKEN_ENCRYPTION_KEY`` is not configured,
@@ -360,7 +368,7 @@ def decrypt_salesforce_token(ciphertext: str, db: Session) -> str:  # noqa: ARG0
 
     key = _salesforce_aes_key()
     try:
-        raw = base64.b64decode(ciphertext[len(_SALESFORCE_AESGCM_PREFIX):])
+        raw = base64.b64decode(ciphertext[len(_SALESFORCE_AESGCM_PREFIX) :])
         nonce, body = raw[:12], raw[12:]
         return AESGCM(key).decrypt(nonce, body, None).decode("utf-8")
     except Exception as exc:
@@ -391,7 +399,9 @@ def _calendly_aes_key() -> bytes:
     return hashlib.sha256(key.encode("utf-8")).digest()
 
 
-def encrypt_calendly_token(plaintext: str, db: Session) -> str:  # noqa: ARG001 - db kept for call-site parity
+def encrypt_calendly_token(
+    plaintext: str, db: Session
+) -> str:  # noqa: ARG001 - db kept for call-site parity
     """Encrypt *plaintext* Calendly OAuth token with AES-256-GCM.
 
     Performed in Python via ``cryptography`` (not pgcrypto SQL — OpenPGP symmetric
@@ -403,10 +413,14 @@ def encrypt_calendly_token(plaintext: str, db: Session) -> str:  # noqa: ARG001 
     key = _calendly_aes_key()
     nonce = os.urandom(12)  # 96-bit nonce, the standard size for AES-GCM
     ciphertext = AESGCM(key).encrypt(nonce, plaintext.encode("utf-8"), None)
-    return _CALENDLY_AESGCM_PREFIX + base64.b64encode(nonce + ciphertext).decode("ascii")
+    return _CALENDLY_AESGCM_PREFIX + base64.b64encode(nonce + ciphertext).decode(
+        "ascii"
+    )
 
 
-def decrypt_calendly_token(ciphertext: str, db: Session) -> str:  # noqa: ARG001 - db kept for call-site parity
+def decrypt_calendly_token(
+    ciphertext: str, db: Session
+) -> str:  # noqa: ARG001 - db kept for call-site parity
     """Decrypt a Calendly OAuth token written by :func:`encrypt_calendly_token`.
 
     Raises ``ValueError`` if ``CALENDLY_TOKEN_ENCRYPTION_KEY`` is not configured or
@@ -417,11 +431,13 @@ def decrypt_calendly_token(ciphertext: str, db: Session) -> str:  # noqa: ARG001
         raise ValueError("ciphertext is empty")
 
     if not ciphertext.startswith(_CALENDLY_AESGCM_PREFIX):
-        raise ValueError("Unrecognized Calendly token ciphertext format (expected gcm1: prefix).")
+        raise ValueError(
+            "Unrecognized Calendly token ciphertext format (expected gcm1: prefix)."
+        )
 
     key = _calendly_aes_key()
     try:
-        raw = base64.b64decode(ciphertext[len(_CALENDLY_AESGCM_PREFIX):])
+        raw = base64.b64decode(ciphertext[len(_CALENDLY_AESGCM_PREFIX) :])
         nonce, body = raw[:12], raw[12:]
         return AESGCM(key).decrypt(nonce, body, None).decode("utf-8")
     except Exception as exc:
@@ -451,7 +467,9 @@ def _ghl_aes_key() -> bytes:
     return hashlib.sha256(key.encode("utf-8")).digest()
 
 
-def encrypt_ghl_token(plaintext: str, db: Session) -> str:  # noqa: ARG001 - db kept for call-site parity
+def encrypt_ghl_token(
+    plaintext: str, db: Session
+) -> str:  # noqa: ARG001 - db kept for call-site parity
     """Encrypt *plaintext* GoHighLevel OAuth token with AES-256-GCM.
 
     Performed in Python via ``cryptography`` (not pgcrypto SQL — OpenPGP symmetric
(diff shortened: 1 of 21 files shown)


app/core/db_encryption.py
```diff
@@ -286,7 +288,7 @@ def decrypt_hubspot_token(ciphertext: str, db: Session) -> str:
     if ciphertext.startswith(_HUBSPOT_AESGCM_PREFIX):
         key = _hubspot_aes_key()
         try:
-            raw = base64.b64decode(ciphertext[len(_HUBSPOT_AESGCM_PREFIX):])
```
codewithwaleed: 🔴 **Blocker: `_SLACK_AESGCM_PREFIX = "gcm1:"` collides with `_CALENDLY_AESGCM_PREFIX = "gcm1:"` already in the file (line 371)**

Both use `"gcm1:"` as their ciphertext prefix. `decrypt_slack_token`'s format guard:
```python
if not ciphertext.startswith(_SLACK_AESGCM_PREFIX):
    raise ValueError("Unrecognized Slack token ciphertext format")
```
…silently passes for any Calendly ciphertext (same prefix), then fails with `"Slack token decryption failed"` due to the wrong key — hiding the real error (wrong function called).

The purpose of these prefixes is exactly to catch cross-function routing bugs. Identical prefixes disable that guard.

Fix:
```python
_SLACK_AESGCM_PREFIX = "slack_gcm1:"
```
No migration cost — the comment confirms there are no existing encrypted Slack rows.

app/services/slack_service.py
```diff
@@ -0,0 +1,536 @@
...
+    the dashboard can surface a clear error rather than silently showing an
+    empty channel list.
+    """
+    access_token = await get_valid_access_token(db, tenant_id)
+    if not access_token:
+        raise ValueError("Slack is not connected for this workspace")
+
+    channels: list[dict] = []
+    cursor: str | None = None
+    while True:
```
codewithwaleed: 🟠 **Major: `list_channels` paginates without an upper bound — O(workspace_channel_count) blocking requests in the admin settings path**

Enterprise Slack workspaces can have 10,000+ channels. With `limit=200` per page, this is 50+ sequential HTTP calls (each a new TCP connection via `_request_with_backoff`), holding the admin `/channels` request for multiple seconds — or triggering 429 rate-limit backoff (1s + 2s + 4s … per page that hits the limit).

The channel-picker UI only needs a few hundred channels. Cap at a small number of pages and signal truncation:

```python
MAX_PAGES = 3  # at most 600 channels; enough for a usable picker
page = 0
while True:
    # ... fetch page ...
    page += 1
    cursor = (payload.get("response_metadata") or {}).get("next_cursor") or None
    if not cursor or page >= MAX_PAGES:
        break
```

Or filter server-side with `exclude_members=true&types=public_channel` to limit scope.

app/services/call_session_service.py
```diff
@@ -77,17 +77,25 @@ async def _enqueue() -> None:
...
 class CallSessionService:
     """Service class for handling call session operations"""
-    
-    def create_call_session(self, db: Session, user_id: uuid.UUID, agent_id: uuid.UUID,
-                           tenant_id: uuid.UUID, twilio_call_sid: str = None,
-                           from_number: str = None, to_number: str = None,
-                           call_type: str = "inbound", assistant_phone_number: str = None,
-                           customer_phone_number: str = None,
-                           session_id: uuid.UUID | None = None,
-                           status: str = "active") -> CallSession:
```
codewithwaleed: 🟠 **Major: `call_session.call_flow` triggers an implicit SQLAlchemy lazy-load — undocumented implicit SELECT in the post-call hot path**

`get_call_session_by_id` loads `CallSession` without eager-loading the `call_flow` relationship. Accessing `.call_flow` here fires an implicit `SELECT * FROM callflow WHERE id = ?` inside the outer try/except block. If the session is in any state where lazy-loading is disabled (expire_on_commit edge case, detached session), this raises a `DetachedInstanceError` that would be caught by the outer `except Exception as slack_exc` and logged as "Slack summary schedule failed" — hiding the real problem.

Be explicit, consistent with the HubSpot/email paths above this block:
```python
if call_session.call_flow_id:
    from app.models.call_flow import CallFlow
    call_flow = db.query(CallFlow).filter(
        CallFlow.id == call_session.call_flow_id
    ).first()
    if call_flow and call_flow.slack_summary_enabled:
        ...
```

app/services/slack_service.py
```diff
@@ -0,0 +1,536 @@
...
+
+_MAX_ERROR_LEN = 500
+
+
+# ─── HTTP with backoff ────────────────────────────────────────────────────────
+
+
+async def _request_with_backoff(method: str, url: str, **kwargs) -> httpx.Response:
+    """Call Slack with exponential backoff on 429 (1s, 2s, 4s, 8s, 16s)."""
+    async with httpx.AsyncClient(timeout=20.0) as client:
```
codewithwaleed: 🟡 **Minor: new `httpx.AsyncClient` per `_request_with_backoff` call — one new TLS handshake per Slack API call**

The `async with httpx.AsyncClient` context manager opens and closes a connection per invocation. For `list_channels` with many pages, each page call to `_request_with_backoff` opens a new connection. The retry loop within one call reuses the connection, but page-to-page doesn't.

For the post-call summary path (a single `chat.postMessage`) this is acceptable. For `list_channels`, pass a shared client or manage it at the call site:
```python
async def list_channels(db, tenant_id):
    ...
    async with httpx.AsyncClient(timeout=20.0) as client:
        while True:
            response = await client.get(
                CONVERSATIONS_LIST_URL,
                headers={"Authorization": f"Bearer {access_token}"},
                params=params,
            )
            ...
```
This also removes the need for `_request_with_backoff` inside the loop (you'd inline the 429 check instead).

app/models/call_flow.py
```diff
@@ -52,6 +52,17 @@ class CallFlow(Base):
...
 
+    # Post Call Actions: post a call summary to a Slack channel after a call
+    # completes. Per-call-flow opt-in — a workspace connecting Slack (see
+    # WorkspaceIntegration, provider="slack") does not activate this for every
+    # call flow. If slack_channel_id is null, the send-time fallback is the
+    # workspace's default_channel_id in WorkspaceIntegration.extra_metadata.
+    slack_summary_enabled = Column(Boolean, default=False, nullable=False, server_default="false")
+    slack_channel_id = Column(String(50), nullable=True)
+    # Denormalized display name of slack_channel_id, so the frontend doesn't
+    # need an extra Slack API round-trip to show the selected channel in the UI.
```
codewithwaleed: 🟡 **Minor: `slack_summary_enabled` is silently skipped for inbound-CRM-sync calls — not documented on the model**

`update_call_session_status` only fires the Slack summary when `not is_inbound_crm_sync_call`. A tenant who enables `slack_summary_enabled=True` on a flow that receives inbound-CRM-sync calls will see no Slack messages posted and have no way to diagnose why — the toggle appears to do nothing.

Add a note here so the model is self-documenting:
```python
# Note: slack_summary_enabled has no effect on inbound-CRM-sync calls
# (call_session.call_metadata["inbound_crm_sync"] is set) — mirrors the
# email summary's is_inbound_crm_sync_call guard in call_session_service.py.
slack_summary_enabled = Column(...)
```

codewithwaleed (changes requested):
## PR #214 Review — Slack post-call summary integration

### Summary
This PR is a well-scoped, well-structured addition of Slack as a post-call-actions delivery channel. The OAuth v2 flow correctly uses a signed JWT state param (carrying `tenant_id` through the redirect), the bot token is AES-256-GCM encrypted at rest using the same pattern as HubSpot/Calendly/GHL, the `send_call_summary` path is correctly fail-open (never blocks call completion), and the `list_channels` admin path correctly raises on failure so the UI shows a real error rather than an empty list. The PR includes 537 lines of tests covering OAuth state, token exchange, channel listing, summary posting, and barge-in interaction.

One security correctness issue and two performance issues require attention before merge.

---

### 🔴 Blockers

**1. `_SLACK_AESGCM_PREFIX = "gcm1:"` collides with `_CALENDLY_AESGCM_PREFIX = "gcm1:"` already in the codebase**

Both `_CALENDLY_AESGCM_PREFIX` (line 371 of the existing `db_encryption.py`) and the new `_SLACK_AESGCM_PREFIX` are set to `"gcm1:"`. This means:

- `decrypt_slack_token` will accept a Calendly ciphertext as input (prefix check passes), then attempt AES-GCM decryption with the Slack key — producing a misleading `ValueError("Slack token decryption failed")` instead of `"Unrecognized ciphertext format"`.
- If future code accidentally passes a Slack token to `decrypt_calendly_token`, the same wrong-key failure occurs with a misleading error.
- The prefix exists to catch "ciphertext routed to the wrong function" bugs — identical prefixes silently disable that guard.

Fix — use a unique prefix:
```python
_SLACK_AESGCM_PREFIX = "slack_gcm1:"
```
(The comment on the new block says "no legacy pgcrypto rows exist", so there's no migration cost.)

---

### 🟠 Major Issues

**2. `list_channels` fetches all pages with no upper bound — blocks the admin request in large workspaces**

```python
while True:
    # ...fetch page...
    cursor = (payload.get("response_metadata") or {}).get("next_cursor") or None
    if not cursor:
        break
```

Slack's `conversations.list` returns every public + private channel in the workspace (up to 1000 per page with `limit=200` = 5 pages minimum, but some enterprise workspaces have 10,000+ channels = 50+ pages). This makes the admin settings load O(workspace_channel_count) with a blocking HTTP fan-out, holding the request for multiple seconds or hitting rate limits that trigger the 429 backoff loop (adding up to 1+2+4+8+16s of sleep).

The dashboard channel-picker only needs the first few hundred channels for selection. Cap the fetch at e.g. 2-3 pages (600 channels) and return with a truncation flag, or use Slack's search API instead:

```python
MAX_CHANNEL_PAGES = 3  # at most 600 channels returned
page = 0
while True:
    # ...
    page += 1
    if not cursor or page >= MAX_CHANNEL_PAGES:
        break
```

**3. `call_session.call_flow` is accessed via SQLAlchemy lazy-load inside the post-call dispatch block**

```python
call_flow = call_session.call_flow
if call_flow and call_flow.slack_summary_enabled:
```

`call_session` was loaded by `get_call_session_by_id` which uses `db.query(CallSession).filter(...).first()` — no `joinedload` or `options(selectinload(...))`. Accessing `.call_flow` triggers an implicit lazy-load SELECT. While this works when the session is open (which it is here), it:

1. Adds an undocumented implicit DB query to the post-call hot path.
2. Creates a subtle dependency on lazy-loading not being disabled (e.g. `expire_on_commit=False` or detached sessions).
3. Is inconsistent with the email summary path directly above it, which queries `call_flow_id` through an explicit lookup.

Fix — be explicit:
```python
if call_session.call_flow_id:
    call_flow = db.query(CallFlow).filter(
        CallFlow.id == call_session.call_flow_id
    ).first()
    if call_flow and call_flow.slack_summary_enabled:
        ...
```

---

### 🟡 Minor Issues

**4. `_request_with_backoff` creates a new `httpx.AsyncClient` per call — one new TCP connection per Slack API call**

```python
async def _request_with_backoff(method, url, **kwargs):
    async with httpx.AsyncClient(timeout=20.0) as client:
        ...
```

Each call to `_request_with_backoff` opens a new connection. For `list_channels` with many pages, this means one new TLS handshake per page. The retry path within a single call reuses the connection (the `async with` wraps the whole retry loop), but adjacent Slack calls in `_send_call_summary_async` (first decrypt check, then postMessage) open separate connections unnecessarily.

For the summary path (just one postMessage), this is acceptable. For `list_channels` with many pages, pass the client in or refactor `list_channels` to manage its own client:

```python
async def list_channels(db, tenant_id):
    ...
    async with httpx.AsyncClient(timeout=20.0) as client:
        while True:
            response = await client.get(CONVERSATIONS_LIST_URL, ...)
            ...
```

**5. Slack summary skipped silently for `is_inbound_crm_sync_call` calls — not documented on the API or model**

```python
if status == "completed" and not is_inbound_crm_sync_call:
    # Slack summary fires here
```

If a call flow has `slack_summary_enabled=True` but the call is an inbound-CRM-sync call, the Slack summary is silently skipped. This is consistent with the email summary behavior above it, but nothing in the `PostCallActionsSettingsUpdate` schema description or the `CallFlow` model comment explains this restriction. A tenant who enables the Slack toggle for an inbound-CRM flow and sees no summaries posted has no way to diagnose why.

Add a note in the `CallFlow` model comment or the API description:
```
# Note: slack_summary_enabled has no effect on inbound CRM sync calls
# (is_inbound_crm_sync_call=True) — mirrors email summary behaviour.
```

---

### Recommendation: **Request Changes** — fix the prefix collision (#1) before merge; #2 and #3 are important correctness/performance issues that should ideally ship with this PR.

Aliict:
Addressed in c675c07:

1. **Blocker (prefix collision)** — fixed. Slack's ciphertext now uses a unique `slack_gcm1:` prefix instead of the shared `gcm1:`.
2. **Unbounded pagination** — fixed. `list_channels` now caps at 3 pages (600 channels) and reuses a single `httpx.AsyncClient` across pages instead of opening one per page. Added a regression test proving the cap holds even when the server keeps returning a `next_cursor`.
3. **Implicit lazy-load** — fixed. The Slack hook now does an explicit `db.query(CallFlow)...` lookup instead of `call_session.call_flow`.
4. **Client-per-call** — covered by #2's client reuse for the multi-call path (`list_channels`); left the single-call `send_call_summary` path as-is since it only makes one Slack request per invocation, matching your note that this is acceptable there.
5. **Undocumented CRM-sync skip** — added a note to both the `CallFlow` model comment and the `slack_summary_enabled` schema field description.

Full Slack + regression suite re-run clean (119 Slack-specific tests, 1519 in the broader services/api suite). Ready for another look.

PR #218, opened 2026-08-23
feat(billing): billing dashboard - plan/credits, Stripe portal, usage reporting, wallet auto-recharge & sharing

Aliict:
## Summary

Follow-up to #217 (already merged) — builds out the full tenant billing-dashboard backend on the same branch:

- `GET /tenants/plan` now also reports monthly price, effective per-minute rate, subscription status, and next billing date.
- `POST /tenants/billing/portal-session` (Stripe billing portal, admin-only), `POST /tenants/plan/cancel` (cancel at period end, audit-logged), `GET /tenants/billing/invoices` (Stripe invoice history).
- Owner-only (workspace-creator-only) cross-workspace usage reporting: `GET /workspace/usage/breakdown(.csv)`, `/recent-activity`, `/minutes-by-month(.csv)` — spend/minutes aggregated across an agency tenant and its direct sub-accounts, gated by a new `require_workspace_owner` dependency that rejects both non-creator admins and sub-account contexts.
- Wallet auto-recharge: tenants can configure a minimum-balance threshold and recharge amount; once `tenant.credits` drops below it, the platform automatically charges their saved Stripe payment method off-session (idempotency-keyed, cooldown-protected via an atomic claim UPDATE verified race-safe against real concurrent Postgres sessions, with a durable trigger trail and a tracked async task so no outcome is silently lost).
- Wallet sharing: a sub-account can be toggled to draw call overage/surcharge credits from its parent tenant's balance instead of its own (its own plan's included-minutes allowance still applies independently), with parent-controlled settings endpoints and an `auto_link_new_workspaces` default for newly created sub-accounts.

## Test plan
- [x] `pytest tests/` — 1268+ passed across affected suites, 0 failed (full-suite run also completed clean earlier in the branch's history)
- [x] Migrations applied and verified against real Postgres (disposable containers), including a dedicated concurrency test proving the auto-recharge cooldown claim is race-safe under genuinely concurrent sessions
- [x] `ruff check .` clean on all touched files
- [x] Every feature went through implementation → code review → fix pass; the two money-moving features (auto-recharge, wallet sharing) got a second review round given real-money risk — findings included and fixed: a Stripe idempotency gap (double-charge risk), an untracked fire-and-forget task, a mid-call wallet-sharing toggle race that could split one call's billing across two tenants, and a balance-leak in the outbound-call 402 response

https://claude.ai/code/session_01VPN1hgjPzsVUTcvCDnR1vW

### DIFF
--- a/alembic/versions/3f67072326de_add_wallet_auto_recharge_trigger_trail.py
+++ b/alembic/versions/3f67072326de_add_wallet_auto_recharge_trigger_trail.py
@@ -1,0 +1,41 @@
+"""add_wallet_auto_recharge_trigger_trail
+
+Revision ID: 3f67072326de
+Revises: 7ec8c07644d0
+Create Date: 2026-08-23 11:20:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision: str = "3f67072326de"
+down_revision: Union[str, Sequence[str], None] = "7ec8c07644d0"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.add_column(
+        "walletautorechargeconfig",
+        sa.Column("last_payment_intent_id", sa.String(), nullable=True),
+    )
+    op.add_column(
+        "walletautorechargeconfig",
+        sa.Column("last_trigger_status", sa.String(), nullable=True),
+    )
+    op.add_column(
+        "walletautorechargeconfig",
+        sa.Column("last_trigger_error", sa.String(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    op.drop_column("walletautorechargeconfig", "last_trigger_error")
+    op.drop_column("walletautorechargeconfig", "last_trigger_status")
+    op.drop_column("walletautorechargeconfig", "last_payment_intent_id")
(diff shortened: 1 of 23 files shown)


codewithwaleed (commented):
## PR #218 Review: Wallet Auto-Recharge + Agency Billing Dashboard

Overall this is a well-structured PR with good attention to idempotency, concurrency safety, and fail-open semantics. The atomic cooldown-claim `UPDATE ... rowcount == 1` pattern for auto-recharge is solid, the `billing_tenant_id` resolution is threaded through the monitoring loop correctly, and the wallet-sharing leak prevention in the 402 error message is a nice catch.

Several issues need attention before merging:

**Critical**
- `credits=50 → credits=0` changed in three places with no explanation — if intentional, needs a migration comment/audit; if accidental, needs reverting
- `_maybe_trigger_wallet_auto_recharge` is called with `tenant_id` (not resolved `billing_tenant_id`), so the auto-recharge config lookup silently misses the parent's config when called from outside the monitoring loop

**Medium**
- `require_workspace_owner` does not guard against `ApiKeyPrincipal` — the dep should explicitly 403 non-User principals before attempting `user_tenant_association` lookups
- SCA/`authentication_required` Stripe errors are silently logged with no user notification path — European cards with required 3DS will accumulate `AUTO-RECHARGE CHARGE FAILED` log lines with no actionable alert to the tenant

**Low**
- `WalletAutoRechargeConfig.updated_at` should be `nullable=False` to match every other timestamp in the file
- `Content-Disposition: attachment; filename=...` missing RFC 6266 quotes around the filename
- `_build_recent_activity` hardcodes `limit(10)` with no pagination — agency with many sub-accounts and high call volume will miss older activity immediately
- `_sums_by_workspace` uses an unbounded `IN (tenant_ids)` — should cap family size or use a subquery to avoid very large IN clauses on agencies with many sub-accounts

PR #221, opened 2026-08-23
feat(voice): visual flow editor backend — JSONB schema, validation, pre-compiled executor

Aliict:
## Summary
- Implements BE6-S6-01: `flow_data`/`compiled_plan` JSONB storage on `call_flow`, graph validation (entry-node/cycle/orphan/outgoing-edge checks), a pre-compiled node executor wired into the live voice pipeline, variable interpolation, and a readonly view mode that strips prompts/phone numbers/webhook URLs from node config.
- Matches the ticket's literal spec: `entry_node_id`/`version`/`compiled_at` as top-level `flow_data` fields, the flat `{node_id: {type, data, next_nodes}}` compiled-plan shape, the `compiled_plan` column name, the `POST /api/v2/flows/{flow_id}/validate` path and its `{valid, errors: [{node_id, message}]}` response, and PUT's `{version, validated}` response shape.
- Fixes a `kb_lookup` node bug where a flow author's chosen `kb_id` was never passed into `rag_service.retrieve()`, so KB-scoped lookups silently searched across every KB bound to the agent instead of the selected one.

## Test plan
- [x] `pytest tests/services/test_flow_graph_service_v2.py tests/api/v2/test_call_flows_visual.py tests/voice/test_flow_executor_v2.py tests/voice/test_flow_pipeline_mixin_v2.py tests/api/v2/test_flow_data_v2.py tests/api/v2/test_flow_data_list.py tests/services/test_rag_service_retrieve.py` — 76 passed
- [x] Full repo suite (`tests/api/`, `tests/voice/`, `tests/services/`, `tests/test_openapi_contract.py`): 2077 passed, 2 failed — both pre-existing, unrelated `test_livekit_recording_service.py` ordering flakes (reproduce in isolation on `main` before this branch's changes too)
- [x] `docs/api/openapi.yaml` regenerated via `scripts/export_openapi.py`
- [x] Alembic migration `20260824_rename_flow_data_compiled_to_compiled_plan` not run against a live DB in this environment (no `TEST_DATABASE_URL` available) — verified via `alembic heads`/chain check only; run `alembic upgrade head` in staging before merging

https://claude.ai/code/session_017fGzwSQprxAijSiVjDSLQ8


codewithwaleed (commented):
## PR #221 Review: Visual Flow Editor Backend — Schema V2

This is a well-scoped refactor: removing the "start node" concept in favour of `entry_node_id` + `greeting` node is cleaner, the compiled plan's flat `{node_id: {type, data, next_nodes}}` lookup is easier to consume than the old sorted-edge list, and the `kb_lookup` node integration is properly dispatched to a thread-pool to avoid blocking the event loop. Test coverage is solid across the new shapes.

Issues to address before merge:

**Correctness**
- `compile_graph`: duplicate `sourceHandle` on two outgoing edges from the same node silently drops the first — no validation error, wrong runtime routing
- `asyncio.get_event_loop()` in an async method is deprecated since Python 3.10 — use `asyncio.get_running_loop()` or the idiomatic `asyncio.to_thread()`

**API contract**
- `GET /{flow_id}/flow-data/validate` removed with no backward-compat alias — frontend must be updated atomically with this deploy or a 404/405 window opens
- The `PUT` response shape changed from `FlowDataResponse` to `FlowDataSaveResponse` (`{version, validated}`) — any frontend reading `flowData` or `flowDataCompiled` from the PUT response will now get `undefined`

**Minor**
- `rag_service.retrieve`: f-string injected into `text()` works but is an anti-pattern; the fragment is safe today but sets a dangerous precedent
- `get_flow_data` readonly path re-runs `validate_graph` on the stripped data on every request — data was already validated at save time, so this is unnecessary overhead

PR #224, opened 2026-08-25
feat(call-flow): add System Webhooks (pre-inbound, dynamic routing, post-call, status)

Aliict:
## Summary

Adds a new **System Webhooks** feature, configured per Call Flow, per the reference product screenshots:

- **Pre-Inbound Call Webhook** — fired before an inbound call connects; POSTs to a tenant URL and injects the `{"variables": {...}}` response into the agent's prompt/greeting via `{{key}}` placeholders. Fail-open: non-2xx/timeout/oversized response → empty variables, call proceeds. No retry (by spec).
- **Dynamic Inbound Call Routing** — toggle; if the Pre-Inbound webhook returns a valid `variables.agent_id`, routes the call to that agent instead of the number's default, with safe fallback on any invalid/missing/cross-tenant value or webhook failure.
- **Post-Call Webhook** — fired after a call ends with the default `{callId, agentId, timestamp, data}` shape, or a tenant-defined custom JSON payload templated against a field catalog (`call_metadata`/`conversation_data`/`transcript`/`analytics`/`header_variables`). Bounded retry (2 attempts, 1min/5min).
- **Status Webhook** — fires on `call.connected` (inbound + outbound), `call.transfer`, `call.ended`. Same bounded retry as Post-Call.

**Architectural change**: inbound Twilio calls previously never resolved or attached a `CallFlow` (`CallSession.call_flow_id` was always `NULL`) — the agent was resolved directly from `PhoneNumber.assistant_id`. This PR closes that gap: inbound calls now resolve the agent's default inbound `CallFlow` and attach it to the session, independent of whether any webhook is configured, so existing inbound-call behavior for numbers with no System Webhooks configured is unchanged.

## Security

- Custom headers (may carry auth cookies/tokens) are pgcrypto-encrypted at rest, reusing the existing `WEBHOOK_SECRET_ENCRYPTION_KEY`/`encrypt_webhook_secret` pattern — no new env var needed.
- All 3 webhook URLs are SSRF-guarded (`assert_public_url`) both at save time and fresh on every delivery attempt (TOCTOU-safe). The Pre-Inbound webhook's SSRF check specifically runs off the event loop (`asyncio.to_thread` + 2s bound) since it's awaited inline on the live Twilio request path — a naive synchronous DNS lookup there could otherwise stall the whole worker.
- Pre-Inbound webhook responses are capped (64KB body / 100 variables / 4000 chars per value) before being stored into `call_session.call_metadata`.
- New endpoints (`PUT /api/v2/flows/{flow_id}/system-webhooks-settings`, `POST /api/v2/flows/{flow_id}/system-webhooks/test`) require admin rank; plaintext headers are never echoed back or logged (audit log only records URLs/toggles/keys).
- Every new query filters by `tenant_id`.

## Database

14 new columns on `CallFlow`, new `SystemWebhookDeliveryLog` table (delivery-outcome log, deliberately excludes headers/params). Migration: `alembic/versions/7e68c9adf64e_add_system_webhooks_to_call_flow.py`. Already applied to the dev DB.

## Test plan

- [x] 115 new tests across templating, encryption, settings endpoint, dispatch service, call-session hooks, inbound routing, and prompt injection — all passing.
- [x] Full suite: 3312 passed / 75 skipped / 0 failed.
- [x] `ruff check .` / `black .` clean.
- [x] Independent code-review pass (multi-tenancy, SSRF, RBAC, async correctness, fail-open guarantees) — 5 findings, all fixed.
- [ ] Manual: configure a System Webhook on a test Call Flow and place a real Twilio call.
- [ ] Manual: same, via Share Demo Link (browser/LiveKit).

https://claude.ai/code/session_01KVjvev3UUetMvYk5gmu8VK


app/services/system_webhook_service.py
```diff
@@ -0,0 +1,824 @@
...
+    try:
+        return decrypt_webhook_headers(ciphertext, db)
+    except ValueError as exc:
+        logger.warning("system_webhook: failed to decrypt stored headers: %s", exc)
+        return {}
+
+
+def _render_headers(headers: dict[str, str], context: dict[str, Any]) -> dict[str, str]:
+    return {
+        render_template(k, context): render_template(v, context)
```
codewithwaleed: **Bug: `_render_headers` applies `render_template` to header *keys* as well as values.**

If a key template like `{{token}}` fails to resolve, it silently becomes an empty string `""`, and you'll be sending an HTTP header with an empty name — which is a protocol error that some HTTP clients accept but others reject or surface as undefined behavior. Template substitution should only be applied to values, not keys:

```python
def _render_headers(headers: dict[str, str], context: dict[str, Any]) -> dict[str, str]:
    return {k: render_template(v, context) for k, v in headers.items()}
```

app/services/system_webhook_service.py
```diff
@@ -0,0 +1,824 @@
...
+            db,
+            tenant_id=call_flow.tenant_id,
+            call_flow_id=call_flow.id,
+            call_session_id=None,
+            webhook_kind="pre_inbound",
+            event_type=None,
+            url=url,
+            headers=headers,
+            query_params=query_params,
+            json_body=None,
```
codewithwaleed: **Design concern: Pre-Inbound webhook fires HTTP POST with an empty body (`json_body=None`).**

`_deliver()` sets `Content-Type: application/json` unconditionally, but `httpx` sends no body when `json=None`. Tenant webhook endpoints expecting a JSON body (e.g. `{"from": "+1...", "to": "+1..."}`) will receive a `POST` with an empty body, which is likely to 400 or 422 on their end.

Consider passing a structured JSON body with the phone-number context that's already available:

```python
json_body={"from": from_number, "to": to_number, **static_metadata},
```

At minimum document clearly in the API spec that the pre-inbound webhook receives an empty POST body.

app/services/system_webhook_service.py
```diff
@@ -0,0 +1,824 @@
...
+    """Build + deliver the Post-Call Webhook payload for one call session.
+    Returns True if delivery succeeded OR nothing was configured (no retry
+    needed either way); False only on an actual delivery failure/timeout.
+    """
+    from app.db.session import SessionLocal
+
+    db: Session = SessionLocal()
+    try:
+        call_session = (
+            db.query(CallSession).filter(CallSession.id == call_session_id).first()
```
codewithwaleed: **Code style: `db.query()` (legacy SQLAlchemy 1.x ORM) used throughout `_dispatch_post_call_webhook` and `_dispatch_status_webhook`.**

Per `CLAUDE.md`: _"SQLAlchemy 2.x: use `select()` + `session.execute()` — not legacy `session.query()`"_. All three queries (`CallSession`, `CallFlow`, `CallLog`) should use the `select()`-based style:

```python
from sqlalchemy import select
call_session = db.execute(
    select(CallSession).where(CallSession.id == call_session_id)
).scalar_one_or_none()
```

Same applies to the `db.query(Agent)` and `db.query(CallFlow)` calls in `voice.py`'s `_resolve_default_inbound_call_flow` and the dynamic routing block.

app/services/system_webhook_service.py
```diff
@@ -0,0 +1,824 @@
...
+            or not call_flow.status_webhook_enabled
+            or not call_flow.status_webhook_url
+        ):
+            return True
+
+        payload: dict[str, Any] = {
+            "callId": str(call_session.id),
+            "apiName": event_type,
+            "apiUrl": None,
+            "status": "success",
```
codewithwaleed: **Bug: status webhook payload hardcodes `"status": "success"` regardless of the actual call outcome.**

The `status` field in the outbound payload will always read `"success"` even when `event_type` is `"call.no_answer"`, `"call.busy"`, or `"call.failed"`. This makes the field meaningless and will mislead downstream consumers.

Either derive this from `event_type` (e.g. `"success"` for `call.ended`, `"failed"` for `call.no_answer`/`call.failed`) or remove the field from the default payload and let tenants control it via custom payload templates.

codewithwaleed (commented):
## System Webhooks PR Review

Solid addition overall — SSRF guard is correctly applied on every delivery attempt (TOCTOU-safe), the fail-open contracts are consistent throughout, and the template-injection safety (response-size cap, variable-count/value-length caps, `{{...}}` distinct from `{...}`) is well thought-out. A few correctness and consistency issues to address:

app/services/system_webhook_service.py
```diff
@@ -0,0 +1,824 @@
...
+            "statusCode": None,
+            "response": None,
+            "timestamp": datetime.now(timezone.utc).isoformat(),
+            "duration": call_session.duration,
+        }
+        if extra:
+            payload.update(extra)
+
+        headers = _render_headers(
```
codewithwaleed: **Bug: `_dispatch_status_webhook` passes the outgoing `payload` dict as the template context for header/query-param rendering.**

The `payload` dict has fields like `callId`, `apiName`, `status`, `timestamp`, `duration` — it does NOT have the `_metadata`, `_system`, or `_variable` namespaces that the pre-inbound and post-call paths expose.

This means header templates like `{{_metadata.apiKey}}` silently resolve to `""` instead of the configured static metadata value. Consistent with the other dispatch functions, build a separate `context` dict:

```python
context = {
    "_system": {"callId": str(call_session.id), "eventType": event_type},
    "_metadata": call_flow.status_webhook_static_metadata or {},
}
headers = _render_headers(
    _decrypt_headers_safe(call_flow.status_webhook_headers_encrypted, db),
    context,
)
```

app/routers/voice.py
```diff
@@ -17,9 +25,12 @@
```
codewithwaleed: **`_resolve_default_inbound_call_flow` uses legacy `db.query()` style.**

Same issue as in `system_webhook_service.py` — should use `select()` + `session.execute()` per the repo's SQLAlchemy 2.x convention.

app/models/system_webhook_log.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import annotations
+
+import uuid
+
+from sqlalchemy import (
```
codewithwaleed: **`SystemWebhookDeliveryLog` is missing `event_type` column in the migration but the model declares it.**

Worth confirming `event_type` is present in `alembic/versions/7e68c9adf64e_add_system_webhooks_to_call_flow.py`. The migration creates the `systemwebhookdeliverylog` table — if `event_type` was added to the model after the migration was generated and the migration wasn't regenerated, `alembic upgrade head` will leave the column missing and cause `IntegrityError` on the first delivery.

app/services/system_webhook_service.py
Aliict: Fixed in f421c2d — `_render_headers` now only renders the header *value*; the key is passed through literally, untouched, so an unresolvable `{{token}}` in a key can no longer produce an empty header name. Left `_render_query_params` templating both key and value as-is — an unresolved/empty query-param name is syntactically valid in a URL (`?=value`), unlike an empty HTTP header name, which is genuinely rejected by many servers/clients, so the asymmetry tracks real protocol risk rather than being an oversight. Added `TestRenderHeaders` with a same-token-in-key-and-value case to prove the divergence is real, not just an updated assertion.

app/services/system_webhook_service.py
Aliict: Fixed in f421c2d — the Pre-Inbound webhook now sends a real JSON body: `{"from": from_number, "to": to_number, "metadata": static_metadata}`. Nested the static metadata under its own `metadata` key rather than spreading it flat (as originally planned) — a follow-up review caught that flat-spreading would let a tenant-configured key literally named `from`/`to` silently clobber the real phone numbers, so nesting avoids that collision entirely. Applied identically to the "Test Webhook" button's pre_inbound branch so the two paths can't diverge. New test asserts the exact `json=` kwarg sent to httpx, plus a dedicated collision test.

app/services/system_webhook_service.py
Aliict: Fixed in f421c2d, and turned out to be one bug wider than flagged. Added `_derive_status_field(event_type, extra)`: for `call.transfer`, derives `status` from `extra["outcome"]`. While fixing that, a follow-up review caught that `call.ended` had the *identical* bug — `CallSessionService.update_call_session_status()` fires `call.ended` for genuinely failed/busy/no-answer terminal calls with no outcome signal at all, so it was also silently always "success". Fixed by threading the real terminal status through: `schedule_status_webhook(call_session.id, "call.ended", extra={"outcome": status})`, and extending `_derive_status_field` to check it for `call.ended` too. `call.connected`/`call.test` remain unconditionally "success" since neither carries an outcome signal today — documented in the function's docstring so that assumption doesn't go stale silently again.

app/services/system_webhook_service.py
Aliict: Fixed in f421c2d — `_dispatch_status_webhook` now builds `context = {"_system": {"callId": ..., "eventType": event_type}}` and renders headers/query-params against that, instead of the outgoing payload dict. Confirmed `CallFlow` has no `status_webhook_static_metadata` column (only `pre_inbound_webhook_static_metadata` exists), so `_metadata`/`_variable` are intentionally omitted from this context rather than inventing a new column — left a code comment explaining why. Applied the same fix to the "Test Webhook" button's status branch. A follow-up review also caught the identical bug in the *post_call* test-branch (rendering against the synthetic test payload instead of the real field-catalog context) — fixed that too, now uses `build_post_call_payload_context` against the tenant's most recent call (or an all-namespaces-empty fallback if they have none yet), matching what the real dispatch path does.

app/routers/voice.py
Aliict: Fixed in f421c2d — `_resolve_default_inbound_call_flow` and the dynamic-routing override `Agent` lookup now use `select()` + `db.execute()`. Kept `.scalars().first()` (not `.scalar_one_or_none()`) for the CallFlow lookup since it filters on non-unique columns and could in principle match >1 row; used `.scalar_one_or_none()` for the Agent lookup since it filters on primary key. Full dynamic-routing test suite (21 tests) passed unchanged before and after.

app/models/system_webhook_log.py
Aliict: Checked — this is a false alarm, no fix needed. `event_type` is present in the migration: `alembic/versions/7e68c9adf64e_add_system_webhooks_to_call_flow.py` line 30, `sa.Column('event_type', sa.String(length=50), nullable=True)`. The migration's already been applied to the dev DB and the column exists there — verified directly against the table's `information_schema.columns`. Thanks for flagging it to double-check though.

app/services/system_webhook_service.py
Aliict: Fixed in f421c2d — `_dispatch_post_call_webhook`/`_dispatch_status_webhook`'s `CallSession`/`CallFlow`/`CallLog` lookups are now `select()` + `db.execute()`. Verified each conversion preserved exact query semantics (`.scalar_one_or_none()` where the filter is on a primary key, so >1 row is impossible; `.scalars().first()` where it isn't) — full regression suite for this feature (147 tests across the affected files) passed unchanged before and after, confirming it's behavior-identical, not just a style change.

PR #236, opened 2026-08-27
Voice pipeline: barge-in fixes, backchannel classifier, and TTS humanization parity

Aliict:
## Summary

- Fixes an interim/final turn-finalization race where `speech_final` could be silently discarded when interim generation failed or returned empty, and encapsulates the interim task's return value to prevent state leaking across turns.
- Removes ElevenLabs bracket-tag/SSML leakage from the Twilio conversational TTS path and aligns Twilio ElevenLabs humanization with the Web Demo's shared pipeline, including a 1.8x clean gain boost for Twilio ElevenLabs outbound audio.
- Disables interim LLM generation on the Twilio transport, removes unused Google TTS precaching, and fixes Deepgram keepalive behavior.
- Introduces `app/voice/backchannel_classifier.py`: a whitelist-first turn classifier (`BARGE_IN` / `SUPPRESS_NON_ACTIONABLE_BACKCHANNEL` / `NORMAL_USER_TURN`) so conversational backchannels ("hey", "yeah", "okay") no longer interrupt active TTS while explicit commands ("stop", "hold on") and genuine short requests still do — replacing a prior word-count-only barge-in heuristic.
- Tunes the single-word barge-in confidence threshold to 0.36 ("fixed barge in" / "fixed two words barge in").
- Adjusts the LLM output token limit ("token limit").

## Test plan

- [x] `tests/test_interim_final_lockout_gate.py` (new, 473 lines) — turn finalization and concurrency isolation regression coverage
- [x] `tests/voice/test_backchannel_classifier.py` (new, 244 lines)
- [x] `tests/voice/test_production_readiness_barge_in.py` (new, 284 lines)
- [x] `tests/services/test_stt_regressions.py` (new, 96 lines)
- [x] `tests/voice/test_twilio_elevenlabs_humanization_and_volume.py` (new, 144 lines)
- [x] `tests/utils/test_eleven_tts_text.py` updated for bracket-tag/SSML leakage fix

### DIFF
--- a/app/routers/bidirectional_stream.py
+++ b/app/routers/bidirectional_stream.py
@@ -366,7 +366,7 @@ class BidirectionalStreamHandler(
         )
         self._twilio_buffer_primed = False  # Track if jitter buffer has been primed
         self._tts_pipeline: TtsPipeline | None = None
-        self._use_ssml = True  # Enable SSML by default
+        self._use_ssml = False  # Disabled by default; only enabled if provider explicitly supports SSML
         # Quick-ack dedup guard: prevent repeated acknowledgements for the same
         # user turn when interim/final regeneration paths overlap.
         self._last_quick_ack_user_norm: str = ""
@@ -487,6 +487,7 @@ class BidirectionalStreamHandler(
             False  # True after first interim triggers LLM for this turn
         )
         self._turn_response_seed_text = ""
+        self._interim_task_response_produced = False
         # In-flight LLM+TTS; wrapped in a task so barge-in can cancel while we await (like dev, but cancelable)
         self._llm_response_task: asyncio.Task | None = None
         self._auto_greeting_sent = False
@@ -655,6 +656,15 @@ class BidirectionalStreamHandler(
                 # Lazy safety net: ensure prompt KB exists for older agents
                 # created before auto-ingest rollout.
                 agent_service.ensure_agent_prompt_ingested(self.db, self.agent)
+                if self.agent:
+                    from app.core.agent_runtime import resolve_tts_runtime
+
+                    runtime = resolve_tts_runtime(self.agent, db=self.db)
+                    # SSML is only supported for non-streaming Google Cloud TTS
+                    self._use_ssml = (
+                        (runtime.adapter_slug or "").lower() == "google"
+                        and not getattr(settings, "VOICE_TTS_STREAMING_ENABLED", True)
+                    )
 
             if self.call_session and self.call_session.call_flow_id:
                 from app.models.call_flow import CallFlow
@@ -920,6 +930,7 @@ class BidirectionalStreamHandler(
         """Stop background LLM+TTS for this turn (barge-in or final regen)."""
         t = self._llm_response_task
         self._llm_response_task = None
+        self._interim_task_response_produced = False
         if t and not t.done():
             t.cancel()
             try:
@@ -995,6 +1006,7 @@ class BidirectionalStreamHandler(
         # ── Phase 1: state check (inside lock, fast) ──────────────────────────
         _should_generate = False
         _need_cancel = False
+        _wait_for_interim_task = None
 
         async with self._llm_turn_serial_lock:
             _now_m = time.monotonic()
@@ -1019,15 +1031,32 @@ class BidirectionalStreamHandler(
                 self._turn_response_started = False
                 self._turn_response_seed_text = ""
                 self._last_interim_text = ""
-                if should_regenerate:
+
+                interim_task = self._llm_response_task
+                if should_regenerate or interim_task is None:
                     _should_generate = True
                     _need_cancel = True
                     self._llm_last_answered_transcript = tstrip
                     self._llm_last_answered_ts = _now_m
+                elif interim_task.done():
+                    interim_res = None
+                    if not interim_task.cancelled() and interim_task.exception() is None:
+                        try:
+                            interim_res = interim_task.result()
+                        except Exception:
+                            interim_res = None
+
+                    if interim_res and str(interim_res).strip():
+                        # Interim LLM task for this turn completed with a valid, committed response.
+                        return
+                    else:
+                        # Interim finished without producing a valid response — generate for final!
+                        _should_generate = True
+                        _need_cancel = True
+                        self._llm_last_answered_transcript = tstrip
+                        self._llm_last_answered_ts = _now_m
                 else:
-                    # Interim LLM already running with the correct transcript —
-                    # let it finish without a second generation.
-                    return
+                    _wait_for_interim_task = interim_task
             else:
                 self._turn_response_seed_text = ""
                 self._last_interim_text = ""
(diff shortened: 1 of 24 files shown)


app/routers/bidirectional_stream.py
```diff
@@ -63,10 +63,8 @@
    - greeting_message is not used on outbound calls
    - Bypasses LLM for that opening only; LLM is not instructed to repeat it verbatim on later "hi"
 
-2. Pre-cached Common Phrases (disabled — implementation commented in code):
-   - 36+ common phrases were pre-generated at connect to warm the MULAW TTS cache
-   - Greetings, acknowledgements, confirmations; <50ms when cache hit
-   - Re-enable by uncommenting asyncio.create_task(self._precache_common_phrases()) and the method
+2. Speculative TTS Prefetch:
+   - Synthesis starts during Deepgram endpointing window, maximising overlap with LLM TTFT.
```
codewithwaleed: **12-second timeout is far too long for a real-time voice call.**

`asyncio.wait_for(asyncio.shield(_wait_for_interim_task), timeout=12.0)` means that if an interim LLM task stalls (API timeout, slow provider), the final STT handler blocks for up to 12 seconds before falling back to generate a fresh response. During this window the caller hears dead silence. For a real-time voice call, 2–3 seconds is the practical maximum before it feels like the call dropped. Reduce to `timeout=2.5` or `3.0` to match the STT endpointing window.

codewithwaleed (commented):
## Review: Barge-In Fixes, Backchannel Classifier & TTS Humanization Parity

The backchannel classifier is well-designed with conservative defaults (unknown phrases are never suppressed). Good fix on the barge-in interim/final deduplication logic. Several issues to address below.

app/utils/audio_utils.py
```diff
@@ -135,7 +135,7 @@ def mix_audio_with_background(
         return tts_audio
 
 
-_VOLUME_MAX_GAIN = 2.0
+_VOLUME_MAX_GAIN = 3.0
```
codewithwaleed: **`_VOLUME_MAX_GAIN = 3.0` will clip 8-bit MULAW audio.**

MULAW is a companded 8-bit codec with a limited dynamic range. Any sample that exceeds the maximum representable amplitude after 3× gain is clipped (hard distortion). The previous 2.0 cap was already aggressive; 3.0 will noticeably clip callers whose input is above ~33% of max amplitude (a normal speaking level for many phones). Either keep 2.0 or add soft-clipping / normalization before encoding.

app/utils/eleven_tts_text.py
```diff
@@ -106,57 +107,58 @@ def _repl(m: re.Match) -> str:
...
+        if key in _ELEVEN_V3_TAG_INNERS:
+            return ""
+        return m.group(0)
+
+    cleaned = _TAG_RE.sub(_repl_audio_tag, cleaned)
+    cleaned = re.sub(r"[ \t]{2,}", " ", cleaned)
+    return cleaned.strip()
 
 
 def supports_elevenlabs_audio_tags(provider_slug: str | None) -> bool:
```
codewithwaleed: **`supports_elevenlabs_audio_tags` is now hardcoded to `return False`, ignoring the `ENABLE_ELEVENLABS_AUDIO_TAGS` setting entirely.**

This is a silent behavioral change: any existing tenant or environment that relied on ElevenLabs audio tags will stop working after this deploy with no config change and no error. The `ENABLE_ELEVENLABS_AUDIO_TAGS` env var is now dead code. If the intent is to disable audio tags by default, flip the default value (`ENABLE_ELEVENLABS_AUDIO_TAGS: bool = False`) rather than hardcoding — that preserves operator opt-in ability and documents the intent clearly.

app/services/gemini_service.py
```diff
@@ -79,14 +79,19 @@ def generate_text(self, prompt: str, system_prompt: str = None,
             if system_prompt:
                 full_prompt = f"System: {system_prompt}\n\nUser: {prompt}"
             
+            # Prepare generation config
+            config_dict = {
+                "temperature": float(temperature),
+                "max_output_tokens": max(int(max_tokens), 1000),
+            }
+            if "2.5" in str(model_name) or "thinking" in str(model_name).lower():
+                config_dict["thinking_config"] = {"thinking_budget": 0}
```
codewithwaleed: **`thinking_config` heuristic is fragile — matches on `"2.5" in model_name` string.**

This check will incorrectly apply `thinking_budget: 0` to Gemini 2.5 models that don't support thinking (e.g. `gemini-2.5-flash-lite`, image-generation variants), potentially causing API errors. It will also miss any future thinking-capable model whose name doesn't contain `"2.5"` or `"thinking"`. Use an explicit allowlist of model IDs that support the thinking API instead:

```python
_THINKING_SUPPORTED_MODELS = {
    "gemini-2.5-pro", "gemini-2.5-flash", "gemini-2.5-pro-preview",
}
if any(model_name.startswith(m) for m in _THINKING_SUPPORTED_MODELS):
    config_dict["thinking_config"] = {"thinking_budget": 0}
```

app/voice/backchannel_classifier.py
```diff
@@ -0,0 +1,261 @@
...
+        "cancel that",
+        "cancel please",
+        "please cancel",
+        "shut up",
+        "be quiet",
+        "quiet please",
+        "listen to me",
+        "listen please",
+        "hang up",
+    }
```
codewithwaleed: **`"hang up"` in `EXPLICIT_COMMAND_PHRASES` is suppressed at low confidence.**

When a caller says `"hang up"` at a confidence below `min_confidence_1w` (0.36), `classify_turn` returns `SUPPRESS_NON_ACTIONABLE_BACKCHANNEL` — the call is not ended and the phrase is silently dropped with no LLM processing. Unlike `"stop"` or `"wait"`, `"hang up"` is an irreversible intent that should always reach the LLM regardless of confidence (or at minimum be logged for review). Consider routing low-confidence `"hang up"` to `NORMAL_USER_TURN` rather than suppressing it.

app/utils/ssml_utils.py
```diff
@@ -10,22 +10,29 @@
...
-    text = re.sub(r'<[^>]+>', '', text)
-    # Remove incomplete SSML tags (tags without closing >, like <break time="150ms)
-    text = re.sub(r'<[^>]*', '', text)
+
+    # Remove complete SSML/XML tags (<tag>content</tag> or <tag/>)
+    text = re.sub(r"<[^>]+>", " ", text)
+    # Remove unclosed tag at end of string (<tag... without closing >)
+    text = re.sub(r"<[^>]*$", " ", text)
+    # Remove trailing unclosed tag fragment at start of string (e.g. rate="90%">)
+    if "<" not in text and ">" in text:
```
codewithwaleed: **New regex `re.sub(r"^[^<]*>", " ", text)` can corrupt legitimate text containing `>` characters.**

The guard `if "<" not in text and ">" in text:` is meant to catch trailing SSML fragments like `rate="90%">`. But if the text is legitimate content like `"items > 5 in stock"` (no `<` present, but has `>`), this regex strips everything up to and including the `>`, leaving `" in stock"`. Any text with a `>` character that arrives in a chunk boundary will be silently corrupted. Consider using a more targeted approach: only strip if the text starts with characters that look like an XML attribute (`[^<]*\s*=\s*["'][^"']*["']\s*>`).

app/services/voice_analysis_service.py
```diff
@@ -321,18 +321,18 @@ def analyze_call_transcript(
                 )
 
                 summary_result = generate_analysis_text(
-                    current_model, current_api_key, summary_prompt, max_tokens=200
+                    current_model, current_api_key, summary_prompt, max_tokens=1000
```
codewithwaleed: **All `max_tokens` values uniformly bumped to 1000 — significant LLM cost increase.**

Summary was 200 tokens, sentiment 150, outcome 120, recommendations 300. All are now 1000. This quadruples to 8× the token budget per post-call analysis, multiplied by every call processed. For tasks like sentiment classification (`positive/negative/neutral`) or outcome detection (a short phrase), 1000 tokens is excessive and wastes API spend. Consider more targeted limits: 400 for summary, 100 for sentiment, 150 for outcome, 500 for recommendations.

PR #237, opened 2026-08-27
Slack post-call summary sync and audio-utils fix

Aliict:
## Summary

- Synchronizes the Slack post-call summary message with the ARQ-scheduled LLM analysis job, and prevents truncated snippets from being posted before the full analysis is ready.
- Small fix in `app/utils/audio_utils.py` (shared MULAW frame helper) picked up incidentally by the humanization work on this branch.

## Test plan

- [x] `tests/services/test_slack_service.py` — expanded coverage for the summary/ARQ sync fix (177 lines changed)
- [x] `tests/utils/test_audio_utils.py` (new, 41 lines)
- [x] `tests/voice/test_livekit_twilio_bridge.py`, `tests/voice/test_rime_eventloop_and_publish_mulaw.py` — updated for the audio_utils change

### DIFF
--- a/app/services/slack_service.py
+++ b/app/services/slack_service.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import asyncio
 import threading
+import time
 import uuid
 from datetime import datetime, timedelta, timezone
 from urllib.parse import urlencode
@@ -70,6 +71,10 @@ _STATE_PURPOSE = "slack_oauth_state"
 _STATE_TTL_MINUTES = 10
 
 _MAX_ERROR_LEN = 500
+
+# Timing constants for synchronizing with background ARQ generate_call_summary job
+_SUMMARY_POLL_INTERVAL_SECONDS = 0.5
+_SUMMARY_MAX_WAIT_SECONDS = 12.0
 
 
 # ─── HTTP with backoff ────────────────────────────────────────────────────────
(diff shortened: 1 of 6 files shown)


codewithwaleed (commented):
## Review: Slack Post-Call Summary Sync & Audio Utils Fix

The G.711 μ-law decoder fix is correct and well-tested — subtracting `ULAW_BIAS` on decode to match the standard ITU-T G.711 spec is the right call. The Slack polling approach for ARQ sync is sensible. A few issues below.

app/services/slack_service.py
```diff
@@ -71,6 +72,10 @@
 
 _MAX_ERROR_LEN = 500
 
+# Timing constants for synchronizing with background ARQ generate_call_summary job
+_SUMMARY_POLL_INTERVAL_SECONDS = 0.5
+_SUMMARY_MAX_WAIT_SECONDS = 12.0
+
 
 # ─── HTTP with backoff ────────────────────────────────────────────────────────
```
codewithwaleed: **`db.refresh(call_session)` is a synchronous SQLAlchemy call inside an `async def` function.**

`db.refresh()` issues a synchronous DB round-trip. Calling it directly from an `async def` blocks the event loop for the duration of the DB query on every poll iteration. Since this polls every 500ms for up to 12 seconds (up to 24 synchronous DB calls), it can stall other coroutines on the event loop for meaningful cumulative time.

Use `asyncio.to_thread(db.refresh, call_session)` to push the blocking call off the event loop, or better, open a dedicated async session for the polling loop.

app/services/slack_service.py
```diff
@@ -402,19 +407,20 @@ async def list_channels(db: Session, tenant_id: uuid.UUID) -> list[dict]:
 def _resolve_channel(
```
codewithwaleed: **`db.query(CallSession)` uses legacy SQLAlchemy 1.x style.**

Per CLAUDE.md: _"SQLAlchemy 2.x: use `select()` + `session.execute()` — not legacy `session.query()`"_.

```python
from sqlalchemy import select
refreshed = db.execute(
    select(CallSession).where(CallSession.id == call_session_id)
).scalar_one_or_none()
```

app/services/slack_service.py
```diff
@@ -22,6 +22,7 @@
 
 import asyncio
 import threading
+import time
 import uuid
```
codewithwaleed: **Bare `except Exception: pass` silently swallows all errors after `ensure_call_summary_locked`.**

The `db.refresh(call_session)` inside the inner `try` block is wrapped in a bare `except Exception: pass`. If the refresh fails (connection error, detached session), the error is swallowed with no log, leaving `call_session` in a potentially stale state before `_sentiment_and_summary` reads from it. Add at minimum `logger.debug("Slack summary: post-ensure refresh failed: %s", exc)`.

app/services/slack_service.py
```diff
@@ -71,6 +72,10 @@
 
 _MAX_ERROR_LEN = 500
```
codewithwaleed: **Polling loop shares the caller's DB session — not safe for concurrent use.**

`_send_call_summary_async` receives `db: Session` from its caller and uses it throughout the polling loop. If `schedule_slack_summary` ever fires this on a session that is also used by other in-flight operations (e.g., in the call-session service), the `db.refresh()` calls inside the loop can interleave with those operations. SQLAlchemy sessions are not safe for concurrent use across coroutines. The polling path should open a fresh `SessionLocal()` scoped to this function.

app/services/slack_service.py
codewithwaleed: **Priority flip removes test coverage for the `transcript_summary`-only fallback.**

The ordering change from `transcript_summary → llm_analysis` to `llm_analysis → transcript_summary` is correct for the stated goal. However, `test_uses_transcript_summary_over_llm_analysis_when_present` was removed rather than updated. There is now no test covering the case where `llm_call_analysis` is present but its `analysis.summary` field is absent or empty, which would fall through to `call_session.transcript_summary`. Add a test case: `llm_call_analysis` present with `summary=None` → should use `transcript_summary`.

